PR #1, opened 2020-03-15
Release guid from removed pods

Mmduh-483:


pkg/daemon/daemon.go
```diff
@@ -234,6 +234,8 @@ func (d *daemon) AddPeriodicUpdate() {
 					glog.Errorf("AddPeriodicUpdate(): failed to update pod annotations with err: %v", err)
 					continue
 				}
+
+				_ = d.guidPool.ReleaseGUID(guidList[index].String())
```
moshe010: if it failed just log a warning the we will know

Mmduh-483: added

pkg/daemon/daemon.go
```diff
@@ -234,6 +234,11 @@ func (d *daemon) AddPeriodicUpdate() {
 					glog.Errorf("AddPeriodicUpdate(): failed to update pod annotations with err: %v", err)
 					continue
 				}
+
+				if err = d.guidPool.ReleaseGUID(guidList[index].String()); err != nil {
+					glog.Warningf("AddPeriodicUpdate(): failed to release guid from removed pod with error: %v",
```
moshe010: please print guid and pod so it will be easy to debug

Mmduh-483: added

PR #3, opened 2020-03-15
Add config structure

Mmduh-483:


pkg/config/config.go
```diff
@@ -1 +1,22 @@
...
+	GuidRangeEnd   string `json:"guid_range_end"`
+}
+
+type SubnetManagerPluginConfig struct {
+	Name        string `json:"subnet_manager_name"`
+	Username    string `json:"subnet_manager_username"`
+	Password    string `json:"subnet_manager_password"`
+	Address     string `json:"subnet_manager_address"`
+	Port        int    `json:"subnet_manager_port"`
+	HttpSchema  string `json:"subnet_manager_httpSchema"`
```
moshe010: what is define here? can we add comment?

pkg/config/config.go
```diff
@@ -1 +1,22 @@
...
+	Certificate string `json:"subnet_manager_certificate"`
```
moshe010: what we expect here? can we had comment?

pkg/config/config.go
```diff
@@ -1 +1,22 @@
...
+	SubnetManager  SubnetManagerPluginConfig
+}
+
+type GuidPoolConfig struct {
+	GuidRangeStart string `json:"guid_range_start"`
+	GuidRangeEnd   string `json:"guid_range_end"`
+}
+
+type SubnetManagerPluginConfig struct {
+	Name        string `json:"subnet_manager_name"`
```
moshe010: it better to hold each plugin config in it own struct.
So you will have UFM config for now in the future there can be more.
so 
UFMConfig
}
in the future :
UFMConfig
OtherPlugin
}

Mmduh-483: > ixed

pkg/config/config.go
Mmduh-483: added

pkg/config/config.go
Mmduh-483: added

pkg/config/config.go
```diff
@@ -1 +1,26 @@
 package config
+
+type DaemonConfig struct {
+	Plugin         int `json:"plugin"`          // Subnet manager plugin name
```
moshe010: Plugin is int?

Mmduh-483: fixed

PR #7, opened 2020-03-16
Implement config pkg

Mmduh-483:


Mmduh-483:
> Please add Yaml which expose the config as env and make sure it works

It is in PR #8 , do you want them to be in the same PR?

Mmduh-483:
> Please add Yaml which expose the config as env and make sure it works

Added

deployment/ib-kubernetes.yaml
```diff
@@ -0,0 +1,132 @@
...
+kind: ConfigMap
+metadata:
+  name: ib-kubernetes-config
+  namespace: kube-system
+data:
+  PLUGIN: "noop"
+  PERIODIC_UPDATE: "5"
+  GUID_RANGE_START: "02:00:00:00:00:00:00:00"
+  GUID_RANGE_END: "02:FF:FF:FF:FF:FF:FF:FF"
+  UFM_USERNAME: ""
```
moshe010: The UFM info should be in a secret and link like this see https://kubernetes.io/docs/concepts/configuration/secret/

Mmduh-483: added

deployment/ib-kubernetes-configmap.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: ib-kubernetes-config
+  namespace: kube-system
+data:
+  PLUGIN: "noop"
```
moshe010: let change to ufm

deployment/ib-kubernetes-ufm-secret.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: ib-kubernetes-ufm-secret
+  namespace: kube-system
+stringData:
+  UFM_USERNAME: ""
+  UFM_PASSWORD: ""
+  UFM_ADDRESS: ""
+  UFM_HTTP_SCHEMA: ""
+  UFM_PORT: ""
+  UFM_CERTIFICATE: ""
```
moshe010: let set it with default values such as: 
  UFM_USERNAME: "admin"
  UFM_PASSWORD: "12345"

pkg/config/config.go
```diff
@@ -1,26 +1,75 @@
...
 
 type GuidPoolConfig struct {
-	GuidRangeStart string `json:"guid_range_start"` // First guid of the pool
-	GuidRangeEnd   string `json:"guid_range_end"`   // Last of the guid pool
+	GuidRangeStart string `json:"guid_range_start" env:"GUID_RANGE_START"` // First guid of the pool
+	GuidRangeEnd   string `json:"guid_range_end" env:"GUID_RANGE_END"`     // Last of the guid pool
 }
 
 type SubnetManagerPluginConfig struct {
 	UFMConfig
```
adrianchiris: for this to be extendable its better to:

```
type SubnetManagerConfig struct {
    Name string 'json:"name"'
    Config json.RawMessage  'json:"config"'
}
```

or

```
type SubnetManagerConfig struct {
    ufm *UFMConfig 'json:"ufm",ommitempty'
}

```

anyway adding it flat is not good imo

pkg/config/config.go
```diff
@@ -1,26 +1,75 @@
...
 }
 
 type UFMConfig struct {
-	Username    string `json:"ufm_username"`    // Username of ufm
-	Password    string `json:"ufm_password"`    // Password of ufm
-	Address     string `json:"ufm_address"`     // IP address or hostname of ufm servere
-	Port        int    `json:"ufm_port"`        // REST API port of ufm
-	HttpSchema  string `json:"ufm_httpSchema"`  // http or https
-	Certificate string `json:"ufm_certificate"` // Certificate of ufm
+	Username    string `json:"ufm_username" env:"UFM_USERNAME"`       // Username of ufm
```
adrianchiris: if we dont add it flat then no need to prefix with `ufm`

pkg/config/config.go
```diff
@@ -1,26 +1,75 @@
...
+	"github.com/caarlos0/env"
+	"github.com/golang/glog"
+)
+
 type DaemonConfig struct {
-	Plugin         string `json:"plugin"`          // Subnet manager plugin name
-	PeriodicUpdate int    `json:"periodic_update"` // Interval between every check for the added and deleted pods
-	GuidPool       GuidPoolConfig
-	SubnetManager  SubnetManagerPluginConfig
+	Plugin         string                    `json:"plugin" env:"PLUGIN"`                   // Subnet manager plugin name
```
adrianchiris: thats the subnet manager plugin right ?
IMO this needs to be partitioned under SubnetManagerPluginConfig
name can be derrived explicitly or implicitly depends on the example i proposed

pkg/config/config.go
```diff
@@ -1,26 +1,75 @@
...
+	PeriodicUpdate int                       `json:"periodic_update" env:"PERIODIC_UPDATE"` // Interval between every check for the added and deleted pods
+	GuidPool       GuidPoolConfig            `json:"guid_pool"`
+	SubnetManager  SubnetManagerPluginConfig `json:"subnet_manager"`
 }
 
 type GuidPoolConfig struct {
-	GuidRangeStart string `json:"guid_range_start"` // First guid of the pool
-	GuidRangeEnd   string `json:"guid_range_end"`   // Last of the guid pool
+	GuidRangeStart string `json:"guid_range_start" env:"GUID_RANGE_START"` // First guid of the pool
```
adrianchiris: no need to prefix with `guid` here IMO

pkg/config/config.go
```diff
@@ -1,26 +1,75 @@
...
+	Password    string `json:"ufm_password" env:"UFM_PASSWORD"`       // Password of ufm
+	Address     string `json:"ufm_address" env:"UFM_ADDRESS"`         // IP address or hostname of ufm server
+	Port        int    `json:"ufm_port" env:"UFM_PORT"`               // REST API port of ufm
+	HttpSchema  string `json:"ufm_http_Schema" env:"UFM_HTTP_SCHEMA"` // http or https
+	Certificate string `json:"ufm_certificate" env:"UFM_CERTIFICATE"` // Certificate of ufm
+}
+
+var supportedPlugins = []string{"noop", "ufm"}
+
+func (dc *DaemonConfig) ReadConfig() error {
```
adrianchiris: where are the []bytes that will be unmarshalled ?

pkg/config/config.go
```diff
@@ -1,26 +1,75 @@
...
+
+	if dc.Plugin == "" {
+		return fmt.Errorf("ValidateConfig(): no plugin selected, supported plugins %v", supportedPlugins)
+	}
+
+	if !dc.isSupportedPlugin() {
+		return fmt.Errorf("ValidateConfig(): not supported plugin %s, supprted plugins %v", dc.Plugin, supportedPlugins)
+	}
+
+	if dc.GuidPool.GuidRangeStart == "" {
```
adrianchiris: do we want to have a default value for start/end range ?

pkg/config/config.go
```diff
@@ -1,26 +1,75 @@
...
+var supportedPlugins = []string{"noop", "ufm"}
+
+func (dc *DaemonConfig) ReadConfig() error {
+	glog.Info("ReadConfig():")
+	return env.Parse(dc)
+}
+
+func (dc *DaemonConfig) ValidateConfig() error {
+	glog.Info("ValidateConfig():")
+	if dc.PeriodicUpdate <= 0 {
```
adrianchiris: do we want to have a default value for it ?

pkg/config/config_test.go
```diff
@@ -0,0 +1,84 @@
...
+	Context("ReadConfig", func() {
+		It("Read configuration from environment variables", func() {
+			dc := &DaemonConfig{}
+
+			Expect(os.Setenv("PLUGIN", "ufm")).ToNot(HaveOccurred())
+			Expect(os.Setenv("PERIODIC_UPDATE", "10")).ToNot(HaveOccurred())
+			Expect(os.Setenv("GUID_RANGE_START", "02:00:00:00:00:00:00:00")).ToNot(HaveOccurred())
+			Expect(os.Setenv("GUID_RANGE_END", "02:00:00:00:00:00:00:FF")).ToNot(HaveOccurred())
+			Expect(os.Setenv("UFM_USERNAME", "admin")).ToNot(HaveOccurred())
+			Expect(os.Setenv("UFM_PASSWORD", "123456")).ToNot(HaveOccurred())
```
adrianchiris: general question : are these cleaned between tests ? in the future it may affect other test cases

README.md
```diff
@@ -6,10 +6,10 @@
...
          * [Building Subnet Manager Plugins](#building-subnet-manager-plugins)
          * [Building Container Image](#building-container-image)
-      * [Configuration reference](#configuration-reference)
+      * [Configuration Reference](#configuration-reference)
       * [Deploying](#deploying)
 
-# IB-Kubernetes
+# InfiniBand Kubernetes
 
 InfiniBand Kubernetes cooperate with [Mellanox InfiniBand SR-IOV CNI](https://github.com/Mellanox/ib-sriov-cni) and
```
adrianchiris: a reword suggestion:
Infiniband kubernetes provides a daemon, 'ib-kubernetes', that works in conjuction with Mellanox Infiniband SR-IOV CNI......... , it acts on kubernetes Pod object changes(Create/Update/Delete), reading the Pod's network annotation and fetching its corresponding network CRD and ......

README.md
```diff
@@ -64,48 +64,27 @@ $ make image
...
+
+**Configurations if "ufm" subnet manager plugin is used for  `deployment/ib-kubernetes-ufm-secret.yaml`:**
+* UFM_USERNAME: Username of UFM. 
+* UFM_PASSWORD: Password of UFM.
+* UFM_ADDRESS: IP address or hostname of UFM server.
+* UFM_HTTP_SCHEMA: http/https, default is https.
+* UFM_PORT: REST API port of UFM default is 443, if `httpSchema` is http then default is 80.
+* UFM_CERTIFICATE: Secure certificate if using secure connection.
 
 ## Deploying
```
adrianchiris: Deployment

pkg/daemon/daemon.go
```diff
@@ -29,6 +31,7 @@ type Daemon interface {
 }
 
 type daemon struct {
+	config     *config.DaemonConfig
```
adrianchiris: why is that a pointer ? its always initialized as zero struct then populated via its methods

pkg/config/config.go
Mmduh-483: @adrianchiris Please check[here](https://github.com/Mellanox/ib-kubernetes/pull/3#discussion_r392674381)

pkg/config/config.go
```diff
@@ -1,26 +1,75 @@
...
+
 type DaemonConfig struct {
-	Plugin         string `json:"plugin"`          // Subnet manager plugin name
-	PeriodicUpdate int    `json:"periodic_update"` // Interval between every check for the added and deleted pods
-	GuidPool       GuidPoolConfig
-	SubnetManager  SubnetManagerPluginConfig
+	Plugin         string                    `json:"plugin" env:"PLUGIN"`                   // Subnet manager plugin name
+	PeriodicUpdate int                       `json:"periodic_update" env:"PERIODIC_UPDATE"` // Interval between every check for the added and deleted pods
+	GuidPool       GuidPoolConfig            `json:"guid_pool"`
+	SubnetManager  SubnetManagerPluginConfig `json:"subnet_manager"`
```
adrianchiris: need to remove the json annotation if not used

pkg/config/config.go
adrianchiris: another suggestion as discussed with @moshe010 is to have all subnet manager configs in the same struct so something like:

```
type SubnetManagerConfig struct {
    PluginName string `env:plugin`
    ufm UFMConfig
}
```

pkg/config/config_test.go
adrianchiris: i recommend to add and AfterEach() clause to clean up env variables

pkg/config/config.go
Mmduh-483: changed

pkg/config/config.go
Mmduh-483: added

README.md
Mmduh-483: changed

README.md
Mmduh-483: fixed

pkg/config/config.go
Mmduh-483: moved

pkg/config/config.go
Mmduh-483: added default values

pkg/config/config.go
Mmduh-483: removed

pkg/daemon/daemon.go
Mmduh-483: changed to be without pointer

pkg/config/config_test.go
Mmduh-483: added after AfterEach()

pkg/config/config.go
moshe010: Why json here?

pkg/config/config.go
moshe010: Add comment why we use json here

pkg/config/config.go
moshe010: Where you define defualt value

pkg/config/config.go
```diff
@@ -1,26 +1,98 @@
...
+	"github.com/caarlos0/env"
+	"github.com/golang/glog"
+)
+
 type DaemonConfig struct {
-	Plugin         string `json:"plugin"`          // Subnet manager plugin name
-	PeriodicUpdate int    `json:"periodic_update"` // Interval between every check for the added and deleted pods
-	GuidPool       GuidPoolConfig
+	PeriodicUpdate int    `env:"PERIODIC_UPDATE"` // Interval between every check for the added and deleted pods
+	RangeStart     string `env:"RANGE_START"`     // First guid of the pool
```
moshe010: Why you move it to be flat?

Mmduh-483: The `GuidPool` is not used any other places, even when pass the startRange and endRange when create the guidPool they are passed as arguments not as struct.

moshe010: So the guid pool should get the config. In the future we may add config and we dont need to to change the signature of the method

Mmduh-483: I will change the newGuidPool to recevie the GuidPoolConfig struct instead of flat args

Mmduh-483: @moshe010 changed

README.md
```diff
@@ -9,15 +9,15 @@
...
+# InfiniBand Kubernetes
 
-InfiniBand Kubernetes cooperate with [Mellanox InfiniBand SR-IOV CNI](https://github.com/Mellanox/ib-sriov-cni) and 
-[Intel Multus CNI](https://github.com/intel/multus-cni), where the InfiniBand Kubernetes read the InfiniBand SR-IOV CNI 
-network attachment configuration created for Multus CNI and read the PKey, to add the newly generated Guids to that PKey,
-for pods with annotation `mellanox.infiniband.app`.
+InfiniBand Kubernetes provides a daemon `'ib-kubernetes`, that works in conjuction with [Mellanox InfiniBand SR-IOV CNI](https://github.com/Mellanox/ib-sriov-cni) and 
+[Intel Multus CNI](https://github.com/intel/multus-cni), it acts on kubernetes Pod object changes(Create/Update/Delete),
+ reading the Pod's network annotation and fetching its corresponding network CRD and and reads the PKey, to add the newly
+ generated Guid or the predefined Guid in `guid` field of CRD `cni-args` to that PKey, for pods with annotation `mellanox.infiniband.app`.
```
moshe010: Why new line

Mmduh-483: removed the new lines

pkg/config/config.go
```diff
@@ -1,26 +1,102 @@
...
+	Address     string `env:"UFM_ADDRESS"`     // IP address or hostname of ufm server
+	Port        int    `env:"UFM_PORT"`        // REST API port of ufm
+	HttpSchema  string `env:"UFM_HTTP_SCHEMA"` // http or https
+	Certificate string `env:"UFM_CERTIFICATE"` // Certificate of ufm
+}
+
+const (
+	defaultPeriodicUpdate = 5
+	defaultRangeStart     = "02:00:00:00:00:00:00:00"
+	defaultRangeEnd       = "02:FF:FF:FF:FF:FF:FF:FF"
```
adrianchiris: a point for thought : 
perhaps reserve two nibbles like in openstack for MAC ?

pkg/config/config.go
```diff
@@ -1,26 +1,102 @@
...
-	PeriodicUpdate int    `json:"periodic_update"` // Interval between every check for the added and deleted pods
+	PeriodicUpdate int `env:"PERIODIC_UPDATE"` // Interval between every check for the added and deleted pods
 	GuidPool       GuidPoolConfig
 	SubnetManager  SubnetManagerPluginConfig
 }
 
 type GuidPoolConfig struct {
-	GuidRangeStart string `json:"guid_range_start"` // First guid of the pool
-	GuidRangeEnd   string `json:"guid_range_end"`   // Last of the guid pool
+	RangeStart string `env:"RANGE_START"` // First guid of the pool
```
adrianchiris: the env variable name is abit cryptic on its own, perhaps GUID_POOL_RANGE_START, GUID_POOL_RANGE_END
i think before it was prefixed by GUID IIRC 

A general guideline could be to name the environment variables like : [component name_]<attribute name> for global daemon attributes we would have the attribute name flat e.g `PERIODIC_UPDATE` (or we could prefix it by "DAEMON" e.g `DAEMON_PERIODIC_UPDATE`)

adrianchiris: and for subnet manager config we can prefix with `SM_`

deployment/ib-kubernetes-ds.yaml
```diff
@@ -0,0 +1,115 @@
...
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: ib-kubernetes
+rules:
+  - apiGroups: [""]
+    resources: ["pods"]
+    verbs: ["get", "list", "patch", "watch"]
+  - apiGroups: ["k8s.cni.cncf.io"]
+    resources: ["*"]
```
adrianchiris: here we just need `NetworkAttachmentDefinition` right ?

can be kept as is, not critical imo...

deployment/ib-kubernetes-ds.yaml
```diff
@@ -0,0 +1,115 @@
...
+        node-role.kubernetes.io/master: ""
+        kubernetes.io/os: "linux"
+      containers:
+        - name: ib-kubernetes
+          image: mellanox/ib-kubernetes
+          imagePullPolicy: IfNotPresent
+          command: ["/usr/bin/ib-kubernetes"]
+          args:
+            - "-logtostderr"
+            - "-v=3"
```
adrianchiris: you set verbose level to 3 but dont use the V style logging in the code.
does that disable the log prints you currently have in code? 

e.g to my understanding you should: glog.V(2).Infof("some log")
then setting -v=2 would make  them visible and -v=3 would hide them

pkg/config/config.go
```diff
@@ -1,26 +1,102 @@
...
+	if dc.SubnetManager.Plugin == "" {
+		return fmt.Errorf("ValidateConfig(): no plugin selected, supported plugins %v", supportedPlugins)
+	}
+
+	if !dc.isSupportedPlugin() {
+		return fmt.Errorf("ValidateConfig(): not supported plugin %s, supprted plugins %v",
+			dc.SubnetManager.Plugin, supportedPlugins)
+	}
+
+	if dc.GuidPool.RangeStart == "" {
```
adrianchiris: no longer needed as you have a default value

pkg/config/config.go
```diff
@@ -1,26 +1,102 @@
...
+		return fmt.Errorf("ValidateConfig(): guids pool start of range not set")
+	}
+
+	if dc.GuidPool.RangeEnd == "" {
```
adrianchiris: no longer needed as you have a default value

pkg/guid/guid_test.go
```diff
@@ -13,40 +14,40 @@ import (
 )
 
 var _ = Describe("GUID Pool", func() {
+	conf := &config.GuidPoolConfig{RangeStart: "02:00:00:00:00:00:00:00", RangeEnd: "02:FF:FF:FF:FF:FF:FF:FF"}
 	Context("NewGuidPool", func() {
 		It("Create guid pool with valid  parameters", func() {
```
adrianchiris: need to add "create guid pool with default values" as well

PR #9, opened 2020-03-16
Add mellanox annotations to network annotations cni-arg

Mmduh-483:


pkg/daemon/daemon.go
```diff
@@ -228,8 +223,19 @@ func (d *daemon) AddPeriodicUpdate() {
 				ibAnnotation = map[string]string{networkName: utils.ConfiguredInfiniBandPod}
 			}
 			ibAnnotationsData, _ := json.Marshal(ibAnnotation)
-
 			pod.Annotations[utils.InfiniBandAnnotation] = string(ibAnnotationsData)
+
+			network := podNetworkMap[pod.UID]
+			(*network.CNIArgs)[utils.InfiniBandAnnotation] = utils.ConfiguredInfiniBandPod
+
+			networks := podNetworkMap[pod.UID]
```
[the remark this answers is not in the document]
Mmduh-483: Fixed and tested

pkg/daemon/daemon.go
```diff
@@ -228,8 +223,19 @@ func (d *daemon) AddPeriodicUpdate() {
 				ibAnnotation = map[string]string{networkName: utils.ConfiguredInfiniBandPod}
 			}
 			ibAnnotationsData, _ := json.Marshal(ibAnnotation)
-
 			pod.Annotations[utils.InfiniBandAnnotation] = string(ibAnnotationsData)
+
+			network := podNetworkMap[pod.UID]
```
moshe010: how this will work if the pod as 2 networks of infiniband? it seem like you will only set the last one

moshe010: Ok it done for the same network which can contains different guid

PR #14, opened 2020-03-18
Fix: Allocating new pod for failed pods in perodic update every loop

Mmduh-483:
When pod failed in perodic update in cases like setting annotations
for the pod or ufm is not unavailable, the daemon allocates new guids
for that pod if it doesn't have guid predefined in the cni-args.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -172,7 +172,7 @@ func (d *daemon) AddPeriodicUpdate() {
 			allocatedGuid, err := utils.GetPodNetworkGuid(network)
 			if err == nil {
 				// User allocated guid manually
-				if err = d.guidPool.AllocateGUID(string(pod.UID)+networkName, allocatedGuid); err != nil {
+				if err = d.guidPool.AllocateGUID(pod.UID, networkName, allocatedGuid); err != nil {
 					glog.Errorf("AddPeriodicUpdate(): %v", err)
 					continue
 				}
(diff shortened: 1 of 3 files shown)


pkg/daemon/daemon.go
```diff
@@ -183,17 +183,25 @@ func (d *daemon) AddPeriodicUpdate() {
 					continue
 				}
 			} else {
-				allocatedGuid, err = d.guidPool.GenerateGUID("")
+				allocatedGuid, err = d.guidPool.GenerateGUID(string(pod.UID) + networkName)
```
moshe010: why Genereate GUID get args?

Mmduh-483: Since GenerateGUID needs to call AllocateGUID, It passes the ID for who is allocating the generated guid.

moshe010: I think we need to change implementation in guid pool.
GenerateGUID  - should just GenerateGUID  and return it. 
and then you need to call AllocateGUID. 
also the AllocateGUID(podID, NetworkID) so you can save it in the map

Mmduh-483: @moshe010  In this PR or another one?

moshe010: In this PR

Mmduh-483: changed

Mmduh-483:
> commit msg it too long and unclear.
> can you elaborate on how this resolve the issue?

Changed

pkg/daemon/daemon.go
```diff
@@ -183,18 +183,30 @@ func (d *daemon) AddPeriodicUpdate() {
...
 				}
 			} else {
-				allocatedGuid, err = d.guidPool.GenerateGUID("")
+				guidAddr, err = d.guidPool.GenerateGUID()
 				if err != nil {
 					glog.Error(err)
 					continue
 				}
+				allocatedGuid = guidAddr.String()
+				if guidErr := d.guidPool.AllocateGUID(string(pod.UID)+networkName, allocatedGuid); guidErr != nil {
```
moshe010: Pass the podid and network each one as arg

Mmduh-483: added

PR #15, opened 2020-03-19
Follow up: Fix validation errors on daemon.go caused by linter

Mmduh-483:


Mmduh-483:
> What linter found this errors? I don't see you enable new linter

It is from #6

Mmduh-483:
> I did not see lint complaning on this can you show me

These issue we

Mmduh-483:
> I did not see lint complaning on this can you show me

It is not issues discovered by linter.
These issues appeared in PR #6 , because in some location we check non related errors.

PR #16, opened 2020-03-19
Fix: Remove guid from pkey in the subnet of the removed pods

Mmduh-483:


pkg/daemon/daemon.go
```diff
@@ -236,10 +237,23 @@ func (d *daemon) AddPeriodicUpdate() {
...
 			finalPassCounter++
 		}
+
+		if ibCniSpec.PKey != "" && len(removedGuidList) != 0 {
+			// Already check the parse above
+			pKey, _ := utils.ParsePKey(ibCniSpec.PKey)
+			if pkeyErr := d.smClient.RemoveGuidsFromPKey(pKey, removedGuidList); pkeyErr != nil {
+				glog.Errorf("AddPeriodicUpdate(): failed to remove guids of removed pods from pKey %s with subnet manager %s with error: %v",
+					ibCniSpec.PKey, d.smClient.Name(), pkeyErr)
+				continue
```
moshe010: why we need  continue?

pkg/daemon/daemon.go
```diff
@@ -236,10 +237,23 @@ func (d *daemon) AddPeriodicUpdate() {
...
 			}
 
 			finalPassCounter++
 		}
+
+		if ibCniSpec.PKey != "" && len(removedGuidList) != 0 {
+			// Already check the parse above
+			pKey, _ := utils.ParsePKey(ibCniSpec.PKey)
+			if pkeyErr := d.smClient.RemoveGuidsFromPKey(pKey, removedGuidList); pkeyErr != nil {
+				glog.Errorf("AddPeriodicUpdate(): failed to remove guids of removed pods from pKey %s with subnet manager %s with error: %v",
```
moshe010: change this to warning

pkg/daemon/daemon.go
Mmduh-483: To prevent removing the network from the map

PR #19, opened 2020-03-21
config: add kubeclient configuration

moshe010:


pkg/config/config.go
```diff
@@ -11,18 +12,23 @@ type DaemonConfig struct {
...
+	RangeStart string `env:"GUID_RANGE_START"` // First guid of the pool
+	RangeEnd   string `env:"GUID_RANGE_END"`   // Last of the guid pool
 }
 
 type SubnetManagerPluginConfig struct {
 	Plugin string `env:"PLUGIN"` // Subnet manager plugin name
 	Ufm    UFMConfig
 }
 
+type KubeConfig struct {
```
adrianchiris: nit: KubeClientConfig ?

im thinking about the next PR that will Prefix each config attribute with its corresponding struct name
and  `KUBE_CLIENT_KUBECONFIG_FILE` seems clearer than `KUBE_KUBECONFIG_FILE`

pkg/config/config.go
```diff
@@ -71,6 +77,15 @@ func (dc *DaemonConfig) ValidateConfig() error {
 		return fmt.Errorf("ValidateConfig(): invalid \"PeriodicUpdate\" value %v", dc.PeriodicUpdate)
 	}
 
+	if os.Getenv("KUBERNETES_SERVICE_HOST") == "" && os.Getenv("KUBERNETES_SERVICE_PORT") == "" {
+		_, err := os.Stat(dc.Kube.File)
```
adrianchiris: you could save a system call by checking if dc.Kube.File is not empty

pkg/config/config_test.go
```diff
@@ -82,6 +82,12 @@ var _ = Describe("Configuration", func() {
 					RangeEnd:   "02:00:00:00:00:00:00:FF"},
 				SubnetManager: SubnetManagerPluginConfig{Plugin: "noop"}}
 
+			os.Setenv("KUBERNETES_SERVICE_HOST", "10.0.0.1")
+			os.Setenv("KUBERNETES_SERVICE_PORT", "443")
+			defer func() {
+				os.Unsetenv("KUBERNETES_SERVICE_HOST")
+				os.Unsetenv("KUBERNETES_SERVICE_PORT")
+			}()
 			err := dc.ValidateConfig()
```
adrianchiris: how does it not fail on the rest of the ValidateConfig calls ?
KUBERNETES_SERVICE_HOST/PORT is not (supposed) to be defined and also KUBE_CONFIG_FILE is not defined in the BeforeEach clause

pkg/config/config.go
```diff
@@ -11,18 +12,23 @@ type DaemonConfig struct {
 	PeriodicUpdate int `env:"PERIODIC_UPDATE"` // Interval between every check for the added and deleted pods
 	GuidPool       GuidPoolConfig
 	SubnetManager  SubnetManagerPluginConfig
+	Kube           KubeConfig
```
adrianchiris: KubeClient?

pkg/daemon/daemon.go
```diff
@@ -42,42 +42,41 @@ type daemon struct {
...
-		glog.Error(err)
+	err = guidPool.InitPool()
+	if err != nil {
 		return nil, err
 	}
 
 	pluginLoader := sm.NewPluginLoader()
-	getSmClientFunc, err := pluginLoader.LoadPlugin(path.Join("/plugins", daemonConfig.SubnetManager.Plugin+".so"),
+
+	workDir, err := os.Getwd()
```
adrianchiris: nit: This, together with path.Join could go in a helper func which accepts subnetManagerConfig and returns the plugin path,
it would be cleaner here plus in the future it will be easier to add an "alternative plugin" base path

pkg/k8s-client/client.go
```diff
@@ -12,7 +12,9 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
-	"sigs.k8s.io/controller-runtime/pkg/client/config"
+	"k8s.io/client-go/tools/clientcmd"
+
+	dconfig "github.com/Mellanox/ib-kubernetes/pkg/config"
```
adrianchiris: IIRC, usually external packages are renamed in case of collision. is there a need to rename here ?
(or was this renamed before removing L15 ?)

pkg/k8s-client/client.go
```diff
@@ -30,25 +32,35 @@ type client struct {
...
+		if err != nil {
+			err = fmt.Errorf("unable to set up client config error %v", err)
+			glog.Error(err)
+			return nil, err
+		}
+	} else {
+		// Try in-cluster config where ib-kubernetes might be running in a kubernetes pod
+		config, err = rest.InClusterConfig()
+		if err != nil {
+			return nil, fmt.Errorf("GetK8sClient: failed to get context for in-cluster kube config: %v", err)
```
adrianchiris: nit: remove "GetK8sClient: "

PR #23, opened 2020-03-22
tls: allow to disable verify hostname

moshe010:
This commit add the option to disable verify hostname in tls connection
setup. Critificate is required when using https. In case verify hostname
is set to false the CN don't have to match the hostname.

The secret yaml was updated with UFM default certificate and ignore
the CN hostname


deployment/ib-kubernetes-ufm-secret.yaml
```diff
@@ -1,12 +1,14 @@
...
...xJcVlnaUxtNW5WRkk0cEdaRDFKcEYyU08vM0FmUlhMYVdQbHZyMUNQSGg2ajVYVjJMU0pyCjRSK25MRjZTTWlidUIvclcwMWVyMGJvaTNFZ3pLZy8yVEtoU1hjSk1WMUpwNGJHRlg0M3YyZGp1NFJXOVVTdFgKRzRPb0pyMUxObkV3TEl5azRoTkhRZzlVdWYvcS9xMStrblE5U09yMExadFN5MDBYeTNHalZIajM4em8veGVqZwpRQW1MbUJpeDFMME94OVRPY3dXcm1WdEQ3QUhWNmlhVkFaMTNPODkwY3ZCNzZkUXNpYmNYTlloeGJzeFNVOUtZCk9BTzhJRDVLSUQxS1V3bnd6Z2hxa29leVQ3Z1hrUFJrbnhHTzVjQ2oyU05jS25jczZvL0JkVGQxMEdlbFFBamwKUTBkSUoyRzk0YVJMendSVjZSUnBzSmx4WUZFQ0F3RUFBYU5RTUU0d0hRWURWUjBPQkJZRUZORXkzK3lUWUdBYgo3MUFYR1U2Z3hQK1pMRmFITUI4R0ExVWRJd1FZTUJhQUZORXkzK3lUWUdBYjcxQVhHVTZneFArWkxGYUhNQXdHCkExVWRFd1FGTUFNQkFmOHdEUVlKS29aSWh2Y05BUUVMQlFBRGdnRUJBQlpxd3RsajBtR3ZKdURublpHelBhT3UKRjVIdWdKL1lVQXM4QnlWanI4NjZMOE9KUWM4czZnbHh3WnR1Y01yaWxkR0JjT0tLeTJjbjJmRHh3T2ZjMnJRdwp3MmU4bVdidllEd0prYStkRmNsWm1EL1l2NU5XNWVIeFc3a3NwWXBjZ0ZZYlFWQVlxeHB1M0t0ZEJtSlduYlRHCkdqak1yZlFXeHZKOENwSi9wUWZZY3NxelVKZ3RBMGMwL0JVN1NlaVplc1pjK2g3RE1yUjYzZUNhYjgveHJWTG8KcVBUQmdCTzVQRmlvNCt0WVgxUUZ0dmRvTE1pRVpwRlRjbG5LNjl2Mlg1Nk90czRXSldLdlRvSWZFWkgzRkhyRQpISHRVT2NoUFN4MUtSTThmS1VUUHVLSUh0VWlQZGlSWkpROERpRHJ6a2k5VWdLazRGRjVIRTAvaG9BWjk1OEE9Ci0tLS0tRU5EIENFUlRJRklDQVRFLS0tLS0K # /etc/pki/tls/certs/ufmlocalhost.crt - base64 format
```
Mmduh-483: why is this here?

deployment/ib-kubernetes-ufm-secret.yaml
```diff
@@ -1,12 +1,14 @@
...
-  name: ib-kubernetes-ufm-secret
   namespace: kube-system
-stringData:
-  UFM_USERNAME: "admin"
-  UFM_PASSWORD: "123456"
-  UFM_ADDRESS: ""
-  UFM_HTTP_SCHEMA: ""
-  UFM_PORT: ""
-  UFM_CERTIFICATE: ""
+  creationTimestamp: null
```
Mmduh-483: Is this needed?

deployment/ib-kubernetes-ufm-secret.yaml
moshe010: This is the base64 of the default certificate.  Now in https you have to provide one

deployment/ib-kubernetes-ufm-secret.yaml
moshe010: no I will remove it

deployment/ib-kubernetes-ufm-secret.yaml
```diff
@@ -1,12 +1,14 @@
 apiVersion: v1
+data:
+  UFM_ADDRESS: ci1kY3M1LTAy # 127.0.0.1 - base64 format
```
adrianchiris: why is UFM address in base 64 format  and not converted in the plugin ? this parameter is user configurable

pkg/drivers/http/client.go
```diff
@@ -33,15 +34,52 @@ type client struct {
...
+					}
+
+					opts := x509.VerifyOptions{
+						Roots:         caCertPool,
+						CurrentTime:   time.Now(),
+						DNSName:       "", // <- skip hostname verification
+						Intermediates: x509.NewCertPool(),
+					}
+
+					for i, cert := range certs {
```
adrianchiris: nit, just for cleaner code but on the expense of creating a new slice for the duration of the loop:

```
for _, cert := range certs[1:] {
    ...
}
```

adrianchiris: if its copy paste code then nvm

pkg/drivers/http/client.go
```diff
@@ -33,15 +34,52 @@ type client struct {
...
+						if i == 0 {
+							continue
+						}
+						opts.Intermediates.AddCert(cert)
+					}
+					_, err := certs[0].Verify(opts)
+					return err
+				},
```
adrianchiris: perhaps worth dealing with the whole  httpClient creation in a separate method

deployment/ib-kubernetes-ufm-secret.yaml
moshe010: 1. you need to cat <cert> | base64 > CERTIFICATE
and use the kubectl create secret generic ib-kubernetes-ufm-secret --from-file=UFM_CERTIFICATE. 
2. we use base64 to avoid special characters user can use none base64 format if he put it under stringData. 
see https://kubernetes.io/docs/concepts/configuration/secret/

deployment/ib-kubernetes-ufm-secret.yaml
moshe010: yes I can move it to stringData

deployment/ib-kubernetes-ufm-secret.yaml
```diff
@@ -1,12 +1,14 @@
...
+  UFM_PASSWORD: MTIzNDU2 # 123456 - base64 format
+  UFM_USERNAME: YWRtaW4= # admin - base64 format
```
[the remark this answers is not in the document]
moshe010: I don't think it will work for the UFM_CERTIFICATE but the other can be moved

pkg/drivers/http/client.go
```diff
@@ -33,15 +34,52 @@ type client struct {
...
-func NewClient(isSecure bool, authMode AuthMode, cert string) Client {
+func NewClient(isSecure bool, authMode AuthMode, cert string, verifyHostName bool) Client {
 	glog.V(3).Info("NewClient():")
 	httpClient := &http.Client{Transport: http.DefaultTransport}
+	caCertPool := x509.NewCertPool()
+	caCertPool.AppendCertsFromPEM([]byte(cert))
 	if isSecure {
-		if cert == "" {
-			httpClient.Transport.(*http.Transport).TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
+		/* #nosec */
```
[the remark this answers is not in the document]
moshe010: because it has the InsecureSkipVerify. We implement the verify peer to not check the hostname but we still need to use InsecureSkipVerify

deployment/ib-kubernetes-ufm-secret.yaml
adrianchiris: README should be updated as well on the process, but this can be another PR

deployment/ib-kubernetes-ufm-secret.yaml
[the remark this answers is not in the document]
adrianchiris: yes i mean except the UFM certificate :) its just for easier definition of the secret.

pkg/drivers/http/client.go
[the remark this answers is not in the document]
adrianchiris: perhaps add a comment before stating : ignore gosec lint check and reason we do it ?

deployment/ib-kubernetes-ufm-secret.yaml
moshe010: Let review the README in another PR

pkg/drivers/http/client.go
moshe010: I want to merge this. let do it as part of code refactoring

PR #26, opened 2020-03-23
Remove passed pods from Add/Delete periodic update

Mmduh-483:


pkg/daemon/daemon.go
```diff
@@ -247,26 +252,27 @@ func (d *daemon) AddPeriodicUpdate() {
...
 			if err != nil {
+				failedPods = append(failedPods, pod)
 				glog.Warningf("AddPeriodicUpdate(): failed to dump networks %+v of pod into json with error: %v",
 					networks, err)
 				continue
 			}
 			pod.Annotations[v1.NetworkAttachmentAnnot] = string(netAnnotations)
 			if err := d.kubeClient.SetAnnotationsOnPod(pod, pod.Annotations); err != nil {
 				if !strings.Contains(strings.ToLower(err.Error()), "not found") {
+					failedPods = append(failedPods, pod)
```
moshe010: if it not found it mean that it was deleted, no need to add it to the failedPods

pkg/daemon/daemon.go
```diff
@@ -247,26 +252,27 @@ func (d *daemon) AddPeriodicUpdate() {
...
 					glog.Errorf("AddPeriodicUpdate(): failed to update pod annotations with err: %v", err)
 					continue
 				}
 
 				if err = d.guidPool.ReleaseGUID(guidList[index].String()); err != nil {
+					failedPods = append(failedPods, pod)
```
moshe010: also here

pkg/daemon/daemon.go
```diff
@@ -376,13 +395,18 @@ func (d *daemon) DeletePeriodicUpdate() {
 			}
 		}
 
-		for _, guidAddr := range guidList {
+		for index, guidAddr := range guidList {
 			if err = d.guidPool.ReleaseGUID(guidAddr.String()); err != nil {
+				failedPods = append(failedPods, passedPods[index])
```
moshe010: what is passedPods? it seem weird that you add the passedPods to the failedPods

pkg/daemon/daemon.go
Mmduh-483: I agree, the case here if it failed and it is not deleted

pkg/daemon/daemon.go
Mmduh-483: The pods that are configured with IB and then passed the remove pkey in delete periodic, Then failed in releasing guid, I am not sure if the is case may happen,
do u want to discard this change?

pkg/daemon/daemon.go
Mmduh-483: this is for removed pods, I agree it shouldn't fail here. removed.

pkg/daemon/daemon.go
moshe010: ha ok I missed the not

PR #27, opened 2020-03-23
plugin: move ufm config to it own plugin

moshe010:


pkg/config/config.go
```diff
@@ -10,28 +10,14 @@ import (
 type DaemonConfig struct {
 	PeriodicUpdate int `env:"PERIODIC_UPDATE" envDefault:"5"` // Interval between every check for the added and deleted pods
 	GuidPool       GuidPoolConfig
-	SubnetManager  SubnetManagerPluginConfig
+	Plugin         string `env:"PLUGIN"` // Subnet manager plugin name
```
adrianchiris: nit: to make things more explicit perhaps SM_PLUGIN ?

pkg/sm/plugin.go
```diff
@@ -13,7 +12,7 @@ import (
 const InitializePluginFunc = "Initialize"
 
 // PluginInitialize is function type that take configurations as []byte and return the subnet manager client.
```
adrianchiris: update comment as well

pkg/config/config.go
```diff
@@ -47,13 +33,13 @@ func (dc *DaemonConfig) ValidateConfig() error {
 		return fmt.Errorf("ValidateConfig(): invalid \"PeriodicUpdate\" value %v", dc.PeriodicUpdate)
 	}
 
-	if dc.SubnetManager.Plugin == "" {
+	if dc.Plugin == "" {
 		return fmt.Errorf("ValidateConfig(): no plugin selected, supported plugins %v", supportedPlugins)
 	}
 
 	if !dc.isSupportedPlugin() {
```
adrianchiris: A general comment: if vendors provide their own plugin do we need to enforce if the plugin is supported?

pkg/config/config.go
adrianchiris: nit: to make things more explicit perhaps rename the env variable to SM_PLUGIN ?

pkg/config/config.go
adrianchiris: If its intended for customers to provide their own, we should not enforce against a fixed list

pkg/config/config.go
moshe010: will update

pkg/config/config.go
moshe010: yes you are right I will remove it

pkg/sm/plugins/ufm/ufm_test.go
```diff
@@ -27,18 +31,23 @@ var _ = Describe("Ufm Subnet Manager Client plugin", func() {
...
 	Context("newUfmPlugin", func() {
 		It("newUfmPlugin ufm plugin", func() {
-			conf := &config.SubnetManagerPluginConfig{Ufm: config.UFMConfig{
-				Username: "admin", Password: "123456", Address: "127.0.0.1", HttpSchema: "http"}}
-			plugin, err := newUfmPlugin(conf)
+			Expect(os.Setenv("UFM_USERNAME", "admin")).ToNot(HaveOccurred())
+			Expect(os.Setenv("UFM_PASSWORD", "123456")).ToNot(HaveOccurred())
+			Expect(os.Setenv("UFM_ADDRESS", "1.1.1.1")).ToNot(HaveOccurred())
+			Expect(os.Setenv("UFM_HTTP_SCHEMA", "http")).ToNot(HaveOccurred())
+			defer os.Clearenv()
```
adrianchiris: the os.Clearenv should be in an AfterEach clause so that when you write a test you dont need to worry about it

pkg/sm/plugins/ufm/ufm_test.go
```diff
@@ -27,18 +31,23 @@ var _ = Describe("Ufm Subnet Manager Client plugin", func() {
...
 			Expect(plugin.Name()).To(Equal("ufm"))
 			Expect(plugin.Spec()).To(Equal("1.0"))
 			Expect(plugin.conf.Port).To(Equal(80))
 		})
 		It("newUfmPlugin with missing address config", func() {
-			conf := &config.SubnetManagerPluginConfig{Ufm: config.UFMConfig{Username: "admin", Password: "123456"}}
-			plugin, err := newUfmPlugin(conf)
+			Expect(os.Setenv("UFM_USERNAME", "admin")).ToNot(HaveOccurred())
+			Expect(os.Setenv("UFM_PASSWORD", "123456")).ToNot(HaveOccurred())
+			Expect(os.Setenv("UFM_HTTP_SCHEMA", "http")).ToNot(HaveOccurred())
```
adrianchiris: which was forgotten here :)

PR #36, opened 2020-03-24
Gracefully terminate daemon goroutines

adrianchiris:
- Extend watcher Run() to return a Stop function to terminate the
  routines that listen to k8s objects
- Modify daemon periodic task execution to use wait.Until()
- Add logic to terminate tasks on os signals SIGINT, SIGTERM in daemon RUN()

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -3,8 +3,11 @@ package daemon
 import (
 	"encoding/json"
 	"net"
+	"os"
+	"os/signal"
 	"path"
 	"strings"
+	"syscall"
 	"time"
 
 	"k8s.io/apimachinery/pkg/types"
@@ -26,7 +29,7 @@ import (
 )
 
 type Daemon interface {
-	// Run run listener for k8s pod events.
+	// Execute Daemon loop, returns when os.Interrupt signal is received
 	Run()
 }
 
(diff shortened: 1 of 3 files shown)


adrianchiris:
Still need to test on a real setup

moshe010:
you have lint failures
pkg/daemon/daemon.go:31:62: `recieved` is a misspelling of `received` (misspell)
	// Execute Daemon loop, returns when os.Interrupt signal is recieved
	                                                            ^
pkg/daemon/daemon.go:122:14: `Recieved` is a misspelling of `Received` (misspell)
		glog.Info("Recieved signal %s. Terminating...", sig)
		           ^
pkg/daemon/daemon.go:120:2: S1000: should use a simple channel send/receive instead of `select` with a single case (gosimple)
	select {
	^
pkg/daemon/daemon.go:118:15: SA1017: the channel used with signal.Notify should be buffered (staticcheck)
	signal.Notify(sigChan, os.Interrupt)
	             ^
Makefile:79: recipe for target 'lint' failed

adrianchiris:
handling SIGTERM signal was added as well as k8s sends that to pods on pod termination to allow them to gracefully shut down.

https://kubernetes.io/docs/concepts/workloads/pods/pod/#termination-of-pods

pkg/daemon/daemon.go
```diff
@@ -101,10 +104,24 @@ func NewDaemon() (Daemon, error) {
...
+	// closing the channel will stop the goroutines executed in the wait.Until() calls below
+	stopPeriodicsChan := make(chan struct{})
+	// Run periodic tasks
+	go wait.Until(d.AddPeriodicUpdate, time.Duration(d.config.PeriodicUpdate)*time.Second, stopPeriodicsChan)
+	go wait.Until(d.DeletePeriodicUpdate, time.Duration(d.config.PeriodicUpdate)*time.Second, stopPeriodicsChan)
 	glog.Info("watcher Run():")
-	d.watcher.Run()
+	// Running Watcher
+	// calling watcherStopFunc() will stop the watcher which listens to events on a separate goroutine
+	watcherStopFunc := d.watcher.Run()
```
moshe010: you are hiding the go routines in the watcher.run which is confusing even with the comment.  Why not go d.watcher.Run()?
also can we use one channel to close all go routines by passing it in the d.watcher.Run?  (and remove the watcherStopFunc)

pkg/watcher/watcher.go
```diff
@@ -12,29 +12,34 @@ import (
...
 	watchList := cache.NewListWatchFromClient(client.GetRestClient(), resource, kapi.NamespaceAll, fields.Everything())
-	return &watcher{eventHandler: eventHandler, watchList: watchList, stopChan: make(chan struct{})}
+	return &watcher{eventHandler: eventHandler, watchList: watchList}
 }
 
-// Run for for infinite listening for k8s resource events
-func (w *watcher) Run() {
+// Run until StopFunc is called, listening for k8s resource events
+func (w *watcher) Run() StopFunc {
+	stopChan := make(chan struct{})
```
moshe010: why get the channel from the caller and let him close it?

pkg/daemon/daemon.go
```diff
@@ -101,10 +104,24 @@ func NewDaemon() (Daemon, error) {
 
 func (d *daemon) Run() {
 	glog.Info("daemon Run():")
-	go wait.Forever(d.AddPeriodicUpdate, time.Duration(d.config.PeriodicUpdate)*time.Second)
-	go wait.Forever(d.DeletePeriodicUpdate, time.Duration(d.config.PeriodicUpdate)*time.Second)
+	// closing the channel will stop the goroutines executed in the wait.Until() calls below
+	stopPeriodicsChan := make(chan struct{})
+	// Run periodic tasks
+	go wait.Until(d.AddPeriodicUpdate, time.Duration(d.config.PeriodicUpdate)*time.Second, stopPeriodicsChan)
+	go wait.Until(d.DeletePeriodicUpdate, time.Duration(d.config.PeriodicUpdate)*time.Second, stopPeriodicsChan)
```
moshe010: defer close(stopPeriodicsChan)

pkg/daemon/daemon.go
moshe010: ok let do this, defer watcherStopFunc()

moshe010: so let change Run to RunInBackground

PR #41, opened 2020-03-25
Set ufm environment variables to optional

Mmduh-483:


adrianchiris:
Also Optional environment variables should be set to `optional`

moshe010:
> Also Optional environment variables should be set to `optional` (those that have default values)

Yes let do this

Mmduh-483:
> Let make all the option with default value as optional

added

PR #43, opened 2020-03-25
Rename ib-kubernetes deployment yaml

adrianchiris:
As we are not using a daemonset the suffix "ds" is missleading

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,5 +86,5 @@ To deploy the InfiniBand Kbubernetes
 ```shell script
 $ kubectl create -f deployment/ib-kubernetes-configmap.yaml
 $ kubectl create -f deployment/ib-kubernetes-ufm-secret.yaml
-$ kubectl create -f deployment/ib-kubernetes-ds.yaml
+$ kubectl create -f deployment/ib-kubernetes.yaml
 ```
--- a/deployment/ib-kubernetes.yaml
+++ b/deployment/ib-kubernetes.yaml
@@ -38,8 +38,11 @@ metadata:
   namespace: kube-system
   annotations:
     kubernetes.io/description: |
-      This daemonset launches the ib-kubernetes daemon for infiniband cni.
+      This deployment launches the ib-kubernetes daemon for Infiniband CNI.
 spec:
+  progressDeadlineSeconds: 600
+  strategy:
+    type: RollingUpdate
   replicas: 1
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -86,5 +86,5 @@ To deploy the InfiniBand Kbubernetes
 ```shell script
 $ kubectl create -f deployment/ib-kubernetes-configmap.yaml
 $ kubectl create -f deployment/ib-kubernetes-ufm-secret.yaml
-$ kubectl create -f deployment/ib-kubernetes-ds.yaml
+$ kubectl create -f deployment/ib-kubernetes-deploy.yaml
```
moshe010: I think ib-kubernetes.yaml should be good enough

deployment/ib-kubernetes-deploy.yaml
```diff
@@ -38,7 +38,7 @@ metadata:
   namespace: kube-system
   annotations:
     kubernetes.io/description: |
-      This daemonset launches the ib-kubernetes daemon for infiniband cni.
+      This deployment launches the ib-kubernetes daemon for infiniband CNI.
```
moshe010: a/infiniband/InfiniBand

deployment/ib-kubernetes-deploy.yaml
```diff
@@ -51,6 +51,10 @@ spec:
     spec:
       serviceAccountName: ib-kubernetes
       hostNetwork: true
+      tolerations:
+        - key: node-role.kubernetes.io/master
+          operator: Exists
+          effect: NoSchedule
```
moshe010: I think we need evaluate what is done in ovn-master yaml and see if it make sense for us [1]
[1] - https://github.com/ovn-org/ovn-kubernetes/blob/master/dist/templates/ovnkube-master.yaml.j2#L41-L65

deployment/ib-kubernetes-deploy.yaml
moshe010: we also should mark it as critical [1]
[1] -https://github.com/ovn-org/ovn-kubernetes/blob/master/dist/templates/ovnkube-master.yaml.j2#L33-L34

adrianchiris:
marking a pod as critical is now done by `priorityClassName` pod attribute and not annotation (since 1.16)

PR #46, opened 2020-03-26
Revert "tls: allow to disable hostanme check"

moshe010:
This reverts commit a8698d70f8af9c4e425547ee0b68a5e04399361d.

The commit reverted because each build of ufm generated new certificates.


pkg/drivers/http/client.go
```diff
@@ -118,7 +75,7 @@ func (c *client) executeRequest(method, url string, expectedStatusCode int, body
 	}
 	resp, err := c.httpClient.Do(req)
 	if err != nil {
-		return nil, fmt.Errorf("failed request %v", err)
+		return nil, fmt.Errorf("faied request %v", err)
```
adrianchiris: failed

pkg/sm/plugins/ufm/ufm_test.go
```diff
@@ -56,33 +56,6 @@ var _ = Describe("Ufm Subnet Manager Client plugin", func() {
 			Expect(err.Error()).To(Equal(`missing one or more required fileds for ufm ["username", "password", "address"]`))
 			Expect(plugin).To(BeNil())
 		})
-		It("newUfmPlugin with missing username config", func() {
```
adrianchiris: should alll of these be removed ?

adrianchiris:
simple revert approving

PR #49, opened 2020-03-29
Add Limitations section

adrianchiris:


adrianchiris:
> The limitation is per node and it in the mlnx driver not in the subnet manager.
I remember a discussion that OpenSM had also some limitations.

if its per node, then it applies also to the openSM node no ?

will the fabric support more than 128pkeys as long as on any single node there are at most 128 pkeys configured ?

adrianchiris:
> What about pkey change in net-crd not supported
There is an open issue on it. i dont think it belongs in the readme

moshe010:
> > The limitation is per node and it in the mlnx driver not in the subnet manager.
> > I remember a discussion that OpenSM had also some limitations.
> 
> @moshe010
> if its per node, then it applies also to the openSM node no ?
> 
> will the fabric support more than 128pkeys as long as on any single node there are at most 128 pkeys configured ?

As far as I member OpenSM don't' have to member in the pkey.

moshe010:
> > What about pkey change in net-crd not supported
> 
> There is an open issue on it. i dont think it belongs in the readme

Ok

adrianchiris:
>As far as I member OpenSM don't' have to member in the pkey.`

OK will update to reflect as per node limitation.

PR #54, opened 2020-04-01
Moved the k8s logic inside guid_pool.go to the daemon.go

### LINKED ISSUE #44: [Improvement] Guid pool should not concern itself with k8s API
Guid pool should not concern itself with k8s API, this should be handled in upper layers.
It should be provided a list of "allocated" GUIDs

Guid pool should not concern itself with which pod got which GUID this should be handled in upper layer.

AbdYsn:
Moved the InitPool() function which read the existing pods guids
and store them from the guid_pool.go to the main.go. Also moved
the guidPodNetworkMap which contains a map between the pods and
their guids from the guid_pool.go to daemon.go and made the necessary
changes.

fixes #44

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -2,6 +2,7 @@ package daemon
 
 import (
 	"encoding/json"
+	"fmt"
 	"net"
 	"os"
 	"os/signal"
@@ -34,11 +35,12 @@ type Daemon interface {
 }
 
 type daemon struct {
-	config     config.DaemonConfig
-	watcher    watcher.Watcher
-	kubeClient k8sClient.Client
-	guidPool   guid.GuidPool
-	smClient   plugins.SubnetManagerClient
+	config            config.DaemonConfig
+	watcher           watcher.Watcher
+	kubeClient        k8sClient.Client
+	guidPool          guid.GuidPool
+	smClient          plugins.SubnetManagerClient
+	guidPodNetworkMap map[string]string // allocated guid mapped to the pod and network
 }
 
 // NewDaemon initializes the need components including k8s client, subnet manager client plugins, and guid pool.
@@ -60,13 +62,9 @@ func NewDaemon() (Daemon, error) {
 		return nil, err
 	}
 
-	guidPool, err := guid.NewGuidPool(&daemonConfig.GuidPool, client)
+	guidPool, err := guid.NewGuidPool(&daemonConfig.GuidPool)
 	if err != nil {
 		return nil, err
-	}
-
-	if guidErr := guidPool.InitPool(); guidErr != nil {
-		return nil, guidErr
 	}
 
 	pluginLoader := sm.NewPluginLoader()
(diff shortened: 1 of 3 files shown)


pkg/daemon/daemon.go
```diff
@@ -426,3 +425,70 @@ func (d *daemon) DeletePeriodicUpdate() {
 
 	glog.Info("DeletePeriodicUpdate(): finished")
 }
+
```
adrianchiris, line 428: nit: I dont think the below should be public

pkg/daemon/daemon.go
```diff
@@ -426,3 +425,70 @@ func (d *daemon) DeletePeriodicUpdate() {
...
+//  InitPool check the guids that are already allocated by the running pods
+func (d *daemon) InitPool() error {
+	glog.Info("InitPool():")
```
adrianchiris, line 431: I believe the correct approach is to let GUID pool track allocated GUIDs by first initializing it with a list of allocated GUIDs then adding/removing GUIDs from the pool as needed.

in that way, there is no need for `guidPodNetworkMap` in the Daemon.
the only issue i see with this approach is with `AllocateGUID()`.

@moshe010 @Mmduh-483 In which flow we risk allocating the same GUID for two pods or two networks on the same pod as depicted in line #472?

moshe010: so user can provide a guid in the pod spec. (like it done with mac address)
In that case we just want to allocate the requested guid if he do not provided we generate the guid.

Mmduh-483: @moshe010 @adrianchiris , In case the guid was allocated then the periodic add failed to do `smPlugin.AddGuidsToPkey` then we need to release the guid again.
The map guidPodNetworkMap was solution for that case, when allocated guid and for some reason the network fail e.g. (adding annotation, setting pkey, ...etc.) then keep it allocated for that guid then check it again in the next periodic update.

moshe010: ok keep it in this PR and we see if can remove it later

adrianchiris: There is a need to keep for each pod circling in the periodic job its state to reduce the number of calls to UFM.
another way of solving that (not related to this commit) is to keep a context for each pod, updating its processing state in the pipeline and skipping steps that where already done.

pkg/daemon/daemon.go
```diff
@@ -434,3 +437,70 @@ func (d *daemon) DeletePeriodicUpdate() {
...
+			}
+		}
+	}
+
+	return nil
+}
+
+func (d *daemon) AllocateGUID(podUid types.UID, networkName, wantedGuid string) error {
+	podNetworkId := string(podUid) + networkName
+	if _, exist := d.guidPodNetworkMap[wantedGuid]; exist {
```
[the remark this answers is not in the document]
adrianchiris: +1

pkg/daemon/daemon.go
adrianchiris: in any case its not related to this commit, as for now we seem to need to have `guidPodNetworkMap`

PR #59, opened 2020-04-02
Use unit64 key instead of string for tracking allocated guids

### LINKED ISSUE #45: [Improvement] GUID pool Work with uint64
This will allow to better optimize the allocation algorithm and reduce memory footprint

Mmduh-483:
fixes #45

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@
 
 InfiniBand Kubernetes provides a daemon `ib-kubernetes`, that works in conjuction with [Mellanox InfiniBand SR-IOV CNI](https://github.com/Mellanox/ib-sriov-cni) and [Intel Multus CNI](https://github.com/intel/multus-cni), it acts on kubernetes Pod object changes(Create/Update/Delete), reading the Pod's network annotation and fetching its corresponding network CRD and and reads the PKey, to add the newly generated Guid or the predefined Guid in `guid` field of CRD `cni-args` to that PKey, for pods with annotation `mellanox.infiniband.app`.
 
+Note: InfiniBand Kubernetes supports x86 architecture.
 ## Subnet Manager Plugins
 
 InifiBand Kubernets uses [Golang plugins](https://golang.org/pkg/plugin/) to communicate with the fabric subnet manager 
--- a/deployment/ib-kubernetes.yaml
+++ b/deployment/ib-kubernetes.yaml
@@ -87,9 +87,6 @@ spec:
           image: mellanox/ib-kubernetes
           imagePullPolicy: IfNotPresent
           command: ["/usr/bin/ib-kubernetes"]
-          args:
-            - "-logtostderr"
-            - "-v=3"
           resources:
             requests:
               cpu: 100m
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -192,7 +192,7 @@ func (d *daemon) AddPeriodicUpdate() {
 			}
 			podNetworkMap[pod.UID] = network
 
-			var guidAddr net.HardwareAddr
+			var guidAddr guid.GUID
 			allocatedGuid, err := utils.GetPodNetworkGuid(network)
 			if err == nil {
 				// User allocated guid manually
@@ -201,7 +201,7 @@ func (d *daemon) AddPeriodicUpdate() {
 					log.Error().Msgf("failed to allocate GUID for pod ID %s, wit error: %v", pod.UID, err)
 					continue
 				}
-				guidAddr, err = net.ParseMAC(allocatedGuid)
+				guidAddr, err = guid.ParseGUID(allocatedGuid)
 				if err != nil {
 					failedPods = append(failedPods, pod)
 					log.Error().Msgf("failed to parse user allocated guid %s with error: %v", allocatedGuid, err)
@@ -238,7 +238,8 @@ func (d *daemon) AddPeriodicUpdate() {
 				pod.Annotations[v1.NetworkAttachmentAnnot] = string(netAnnotations)
 			}
 
-			guidList = append(guidList, guidAddr)
+			// used GUID as net.HardwareAddress to use it in sm plugin which receive n[]et.HardwareAddress as parameter
+			guidList = append(guidList, guidAddr.HardWareAddress())
 			passedPods = append(passedPods, pod)
 		}
 
--- a/pkg/guid/guid.go
+++ b/pkg/guid/guid.go
@@ -1,0 +1,41 @@
+package guid
+
+import (
+	"fmt"
+	"net"
+)
+
+// GUID address is an uint64 encapsulation for network hardware address
+type GUID uint64
+
+const guidLength = 8
+
+// ParseGUID parses string only as GUID 64 bit
+func ParseGUID(s string) (GUID, error) {
+	ha, err := net.ParseMAC(s)
+	if err != nil {
+		return 0, err
+	}
+	if len(ha) != guidLength {
+		return 0, fmt.Errorf("invalid GUID address %s", s)
+	}
+
+	return GUID(uint64(ha[0])<<56 | uint64(ha[1])<<48 | uint64(ha[2])<<40 | uint64(ha[3])<<32 | uint64(ha[4])<<24 |
+		uint64(ha[5])<<16 | uint64(ha[6])<<8 | uint64(ha[7])), nil
+}
+
+// String returns the string representation of GUID
+func (g GUID) String() string {
+	return g.HardWareAddress().String()
+}
+
+func (g GUID) HardWareAddress() net.HardwareAddr {
+	value := uint64(g)
+	ha := make(net.HardwareAddr, guidLength)
+	for idx := 7; idx >= 0; idx-- {
+		ha[idx] = byte(value & 0xFF)
+		value >>= 8
+	}
+
+	return ha
+}
--- a/pkg/guid/guid_pool.go
+++ b/pkg/guid/guid_pool.go
@@ -2,9 +2,6 @@ package guid
 
 import (
 	"fmt"
-	"net"
-	"regexp"
-	"strings"
 
 	"github.com/Mellanox/ib-kubernetes/pkg/config"
 	k8s_client "github.com/Mellanox/ib-kubernetes/pkg/k8s-client"
@@ -26,7 +23,7 @@ type GuidPool interface {
 	// It returns the allocated guid or error if range is full.
 	AllocateGUID(types.UID, string, string) error
 
-	GenerateGUID() (net.HardwareAddr, error)
+	GenerateGUID() (GUID, error)
 
 	// ReleaseGUID release the reservation of the guid.
 	// It returns error if the guid is not in the range.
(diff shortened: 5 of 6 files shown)


pkg/guid/guid_pool.go
```diff
@@ -169,55 +176,65 @@ func (p *guidPool) ReleaseGUID(guid string) error {
...
+	var result uint64 = 0
+	currentFactor := uint64(math.Pow(0x100, 7))
 	for idx := 0; idx <= 7; idx++ {
-		if startGUID[idx] < endGUID[idx] {
-			return nil
-		}
-		if startGUID[idx] > endGUID[idx] {
-			return fmt.Errorf("invalid guid range. rangeStart: %s rangeEnd: %s", startGUID.String(), endGUID.String())
-		}
+		result += currentFactor * uint64(guid[idx])
```
moshe010, line 222: why not do it like this https://github.com/cilium/cilium/blob/master/pkg/mac/mac.go#L52 but in a for loop.
I mean use shift operator<<

pkg/guid/guid_pool.go
```diff
@@ -169,55 +176,65 @@ func (p *guidPool) ReleaseGUID(guid string) error {
...
-func checkGuidRange(startGUID, endGUID net.HardwareAddr) error {
-	glog.V(3).Infof("checkGuidRange(): startGUID %v, endGUID %s", startGUID.String(), endGUID.String())
+// isValidGUID check if the guild is valid
+func isValidGUID(guid string) bool {
+	match, _ := regexp.MatchString("^([0-9a-fA-F]{2}:){7}[0-9a-fA-F]{2}$", guid)
+	return match && guid != "00:00:00:00:00:00:00:00" && !strings.EqualFold(guid, "ff:ff:ff:ff:ff:ff:ff:ff")
+}
+
+// guidToUint64 convert guid address to uint64 number
+func guidToUint64(guid net.HardwareAddr) uint64 {
```
moshe010, line 218: I think it better to encapsulate the net.HardwareAddr. 
I mean create a guid.go

which has 
in the package level (not method)
ParseMAC (call net.ParseMAC + check it has 8 bytes)
String ( call net.String)
UInt( out implementation) 
then you can remove the isValidGUID

pkg/guid/guid_pool.go
```diff
@@ -124,43 +114,43 @@ func (p *guidPool) GenerateGUID() (net.HardwareAddr, error) {
...
-			if p.currentGUID.String() == p.rangeEnd.String() {
-				break
-			}
-			p.currentGUID = p.getNextGUID(p.currentGUID)
+	rangeEndIndex := p.rangeEnd.Uint64()
+	for guid := p.currentGUID; guid.Uint64() <= rangeEndIndex; guid = guid.nextGUID() {
+		guidIndex := guid.Uint64()
+		if _, ok := p.guidPoolMap[guidIndex]; !ok {
+			p.currentGUID = guid.nextGUID()
+			return net.HardwareAddr(guid), nil
```
moshe010, line 122: why return HardwareAddr and not GUID?

pkg/guid/guid_pool.go
```diff
@@ -35,45 +33,37 @@ type GuidPool interface {
...
 	kubeClient        k8s_client.Client // kubernetes client
-	rangeStart        net.HardwareAddr  // first guid in range
-	rangeEnd          net.HardwareAddr  // last guid in range
-	currentGUID       net.HardwareAddr  // last given guid
-	guidPoolMap       map[string]bool   // allocated guid map and status
+	rangeStart        GUID              // first guid in range
+	rangeEnd          GUID              // last guid in range
+	currentGUID       GUID              // last given guid
+	guidPoolMap       map[uint64]bool   // allocated guid map and status
 	guidPodNetworkMap map[string]string // allocated guid mapped to the pod and network
```
moshe010, line 40: so this guidPodNetworkMap  will be remove in the other PR right?

pkg/guid/guid.go
```diff
@@ -0,0 +1,47 @@
...
+	return net.HardwareAddr(g).String()
+}
+
+func (g GUID) Uint64() uint64 {
+	return uint64(g[0])<<56 | uint64(g[1])<<48 | uint64(g[2])<<40 | uint64(g[3])<<32 |
+		uint64(g[4])<<24 | uint64(g[5])<<16 | uint64(g[6])<<8 | uint64(g[7])
+}
+
+// nextGUID return next guid
+func (g GUID) nextGUID() GUID {
```
moshe010, line 35: why it not a private method in the guildpool?

pkg/guid/guid_pool.go
```diff
@@ -124,43 +114,43 @@ func (p *guidPool) GenerateGUID() (net.HardwareAddr, error) {
...
 		}
+	}
 
-		copy(p.currentGUID, p.rangeStart)
+	rangeCurrentIndex := p.rangeStart.Uint64()
+	for guid := p.rangeStart; guid.Uint64() <= rangeCurrentIndex; guid = guid.nextGUID() {
+		guidIndex := guid.Uint64()
+		if _, ok := p.guidPoolMap[guidIndex]; !ok {
+			p.currentGUID = guid.nextGUID()
+			return net.HardwareAddr(guid), nil
```
moshe010, line 131: same as above

pkg/guid/guid_pool.go
Mmduh-483: It should be in #54

pkg/guid/guid_pool.go
Mmduh-483: changed to GUID

pkg/guid/guid.go
Mmduh-483: moved

pkg/guid/guid_pool.go
Mmduh-483: moved

pkg/daemon/daemon.go
```diff
@@ -248,7 +248,7 @@ func (d *daemon) AddPeriodicUpdate() {
 				pod.Annotations[v1.NetworkAttachmentAnnot] = string(netAnnotations)
 			}
 
-			guidList = append(guidList, guidAddr)
+			guidList = append(guidList, net.HardwareAddr(guidAddr))
```
moshe010, line 251: why not GUID?

Mmduh-483: @moshe010  Do you prefer to change all the net.HardwareAddress to GUID, including the sm plugin parameters?

moshe010: I see just add a comment why you add it  as net.HardwareAddr. and we change it in another PR

pkg/daemon/daemon.go
```diff
@@ -248,7 +248,8 @@ func (d *daemon) AddPeriodicUpdate() {
 				pod.Annotations[v1.NetworkAttachmentAnnot] = string(netAnnotations)
 			}
 
-			guidList = append(guidList, guidAddr)
+			// used GUID as net.HardwareAddress to use it in sm plugin which receive n[]et.HardwareAddress as parameter
```
moshe010, line 251: the comment is not clear also what is the []
return net.HardwareAddress has other modules such as sm plugin still use it.
We should fix this in another PR

Mmduh-483: Apart from `guid` module, the `net.HardwareAddress` is used in `daemon`, `ib-utils`, and `sm plugins`.

Mmduh-483: @moshe010 updated the comment

pkg/guid/guid.go
```diff
@@ -0,0 +1,32 @@
+package guid
+
+import (
+	"fmt"
+	"net"
+)
+
+// GUID address is an net.HardwareAddr encapsulation
+type GUID net.HardwareAddr
```
adrianchiris, line 9: i see alot of conversions to Uint64() in guid pool,

perhaps the type should be uint64 and the package would have a `func FromHardwareAddress(address net.HardwareAddr) GUID` ?

also add `GUID.HardwareAddress()` method for the other way around to be used in the daemon

pkg/guid/guid_pool.go
```diff
@@ -169,55 +148,55 @@ func (p *guidPool) ReleaseGUID(guid string) error {
...
 }
 
-func (p *guidPool) getNextGUID(currentGUID net.HardwareAddr) net.HardwareAddr {
+func isValidRange(rangeStart, rangeEnd GUID) bool {
+	return rangeStart.Uint64() <= rangeEnd.Uint64() && rangeStart.Uint64() != 0 && rangeEnd.Uint64() != 0xFFFFFFFFFFFFFFFF
+}
+
+// nextGUID return next guid
+func nextGUID(guid GUID) GUID {
+	nextGUID := make(GUID, len(guid))
```
adrianchiris, line 177: if GUID was a uint64 then you would return : guid + 1 (see my above suggestion)

pkg/guid/guid.go
moshe010: yes make seen to me lates do it.

moshe010:
also fix the merge confilct

pkg/guid/guid.go
```diff
@@ -0,0 +1,41 @@
+package guid
+
+import (
+	"fmt"
+	"net"
+)
+
+// GUID address is an net.HardwareAddr encapsulation
```
adrianchiris, line 8: the comment is not accurate

Mmduh-483: fixed

pkg/guid/guid.go
```diff
@@ -0,0 +1,41 @@
...
+func ParseGUID(s string) (GUID, error) {
+	ha, err := net.ParseMAC(s)
+	if err != nil {
+		return 0, err
+	}
+	if len(ha) != guidLength {
+		return 0, fmt.Errorf("invalid GUID address %s", s)
+	}
+
+	return GUID(uint64(ha[0])<<(7*guidLength) | uint64(ha[1])<<(6*guidLength) | uint64(ha[2])<<(5*guidLength) | uint64(ha[3])<<(4*guidLength) |
```
adrianchiris, line 23: can you do this in a for loop ?

Also i dont think you need to do this for big-endian CPU. do we support ppc64 with big endian architecture ?

pkg/guid/guid.go
```diff
@@ -0,0 +1,41 @@
...
+// String returns the string representation of GUID
+func (g GUID) String() string {
+	return g.HardWareAddress().String()
+}
+
+func (g GUID) HardWareAddress() net.HardwareAddr {
+	value := uint64(g)
+	ha := make(net.HardwareAddr, guidLength)
+	for idx := 7; idx >= 0; idx-- {
+		ha[idx] = byte(value % 0x100)
```
adrianchiris, line 36: please use bitwise operations

adrianchiris: endianness should be taken into account if this will run on big endian architecture (e.g ppc64 big endian)

pkg/guid/guid.go
moshe010: currently we don't support ppc64 so this is fine

moshe010: > can you do this in a for loop ?
I don't think it important for now or should block the patch

> 
> Also i dont think you need to do this for big-endian CPU. do we support ppc64 with big endian architecture ?

pkg/guid/guid.go
moshe010: I agree change it to bitwise operation. Regarding the big endian we done in the future per arch.
Also add in the redame that we support x86 arch only

pkg/guid/guid.go
adrianchiris: > currently we don't support ppc64 so this is fine

that what blocked the patch :)

moshe010: yes :)

README.md
```diff
@@ -19,6 +19,7 @@
 
 InfiniBand Kubernetes provides a daemon `ib-kubernetes`, that works in conjuction with [Mellanox InfiniBand SR-IOV CNI](https://github.com/Mellanox/ib-sriov-cni) and [Intel Multus CNI](https://github.com/intel/multus-cni), it acts on kubernetes Pod object changes(Create/Update/Delete), reading the Pod's network annotation and fetching its corresponding network CRD and and reads the PKey, to add the newly generated Guid or the predefined Guid in `guid` field of CRD `cni-args` to that PKey, for pods with annotation `mellanox.infiniband.app`.
 
+Note: InfiniBand Kubernetes supportes x86 architecture.
```
adrianchiris, line 22: nit: supportes -> supports

deployment/ib-kubernetes.yaml
```diff
@@ -87,9 +87,6 @@ spec:
           image: mellanox/ib-kubernetes
           imagePullPolicy: IfNotPresent
           command: ["/usr/bin/ib-kubernetes"]
-          args:
-            - "-logtostderr"
```
adrianchiris, line 91: @Mmduh-483  remnants of glog ?

pkg/guid/guid.go
```diff
@@ -0,0 +1,41 @@
...
+func (g GUID) String() string {
+	return g.HardWareAddress().String()
+}
+
+func (g GUID) HardWareAddress() net.HardwareAddr {
+	value := uint64(g)
+	ha := make(net.HardwareAddr, guidLength)
+	for idx := 7; idx >= 0; idx-- {
+		ha[idx] = byte(value & 0xFF)
+		value >>= guidLength
```
adrianchiris, line 37: you are shifting a byte's bit length right ? 
even though their value match i would prefer not to rely on it so either explicitly do >> 8 or have a const for it.

deployment/ib-kubernetes.yaml
Mmduh-483: yes

pkg/guid/guid.go
Mmduh-483: fixes

Issue #57, opened 2020-04-02
Travis CI is passing even when unit test fails


Mmduh-483:
@adrianchiris @moshe010 Check CI job of #55 , it is passing although the unit test is failing for `PluginLoader`

PR #60, opened 2020-04-03
make: improve Makefile

moshe010:


Mmduh-483:
fixes #57

Mmduh-483:
mocks pkgs shouldn't appear in the results of coverage tests, check [here](https://travis-ci.org/github/Mellanox/ib-kubernetes/builds/670400192#L477).

In the commit message there is a typo in `aliase`:
4. remove aliase check and tests aliases

moshe010:
> mocks pkgs shouldn't appear in the results of coverage tests, check [here](https://travis-ci.org/github/Mellanox/ib-kubernetes/builds/670400192#L477).
> 
> In the commit message there is a typo in `aliase`:
> 4. remove aliase check and tests aliases
right, but the mock folder don't tests so it don't affect the coverage. Unfortunately we have also pkgs which don't have test so the code coverage is much lower than what is reported

Thanks, I will fix the commit msg

PR #61, opened 2020-04-05
Fix ufm secret file

Mmduh-483:
[root@r-cloudx2-01 ib-kubernetes]# kubectl create -f deployment/ib-kubernetes-ufm-secret.yaml 
error: error validating "deployment/ib-kubernetes-ufm-secret.yaml": error validating data: ValidationError(Secret): unknown field "string" in io.k8s.api.core.v1.Secret; if you choose to ignore these errors, turn validation off with --validate=false


Mmduh-483:
@moshe010

PR #62, opened 2020-04-05
modules: update network-attachment-definition-client

moshe010:
Update network-attachment-definition-client that uses newer kubernetes


Mmduh-483:
@moshe010 tested, it is working fine

PR #64, opened 2020-04-08
Fix allocating guid

### LINKED ISSUE #58: Networks can allocated GUID out of pool range
If the range is `02:00:00:00:00:00:00:00` - `02:00:00:00:00:00:00:FF`, Networks can allocated  GUID like `02:00:00:00:00:FF:00:00`, which is out of pool range.

Mmduh-483:
fixes #58

### DIFF
--- a/pkg/guid/guid_pool.go
+++ b/pkg/guid/guid_pool.go
@@ -139,6 +139,10 @@ func (p *guidPool) AllocateGUID(podUid types.UID, networkId, guid string) error 
 		return err
 	}
 
+	if guidAddr < p.rangeStart || guidAddr > p.rangeEnd {
+		return fmt.Errorf("out of range guid %s, pool range %v - %v", guid, p.rangeStart, p.rangeEnd)
+	}
+
 	podNetworkId := string(podUid) + networkId
 	if _, exist := p.guidPoolMap[guidAddr]; exist {
 		if podNetworkId != p.guidPodNetworkMap[guid] {
--- a/pkg/guid/guid_test.go
+++ b/pkg/guid/guid_test.go
@@ -100,8 +100,8 @@ var _ = Describe("GUID Pool", func() {
 	})
 	Context("GenerateGUID", func() {
 		It("Generate guid when range is not full", func() {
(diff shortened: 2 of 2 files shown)


Mmduh-483:
> Please add the Fixes #

Already added


> Also unit test are failing

Fixed

pkg/guid/guid.go
```diff
@@ -0,0 +1,32 @@
+package guid
+
+import (
+	"fmt"
+	"net"
+)
+
+// GUID address is an net.HardwareAddr encapsulation
+type GUID net.HardwareAddr
```
adrianchiris, line 9: perhaps type it as : InfinibandGUID ?
the term GUID is very general.

adrianchiris, line 9: i see alot of conversions to Uint64() in guid pool,

perhaps the type should  be `uint64` and the package would have a `func FromHardwareAddress(address net.HardwareAddr) GUID` ?

also add `GUID.HardwareAddress()` method for the other way around to be used in the daemon

pkg/guid/guid_pool.go
```diff
@@ -169,55 +148,60 @@ func (p *guidPool) ReleaseGUID(guid string) error {
...
 	return nil
 }
 
-func (p *guidPool) getNextGUID(currentGUID net.HardwareAddr) net.HardwareAddr {
+func isValidRange(rangeStart, rangeEnd GUID) bool {
+	return rangeStart.Uint64() <= rangeEnd.Uint64() && rangeStart.Uint64() != 0 && rangeEnd.Uint64() != 0xFFFFFFFFFFFFFFFF
+}
+
+// nextGUID return next guid
+func nextGUID(guid GUID) GUID {
```
adrianchiris, line 181: if GUID was a uint64 then you would return : guid + 1 (see my above suggestion)

Mmduh-483:
> the second commit is related to PR#59 i dont understand why its here.

This PR request #59 , It is just the second commit should be reviewed

moshe010:
@Mmduh-483 , maybe next time indicated that other PR need to be review before this one

adrianchiris:
IMO this could have been merged without the dependency

moshe010:
> IMO this could have been merged without the dependency

He is using the uint64 change to solve i, but anyway not important for now

PR #67, opened 2020-04-14
Fix: daemon crash on adding new pods due to some fields not initialized

Mmduh-483:


Mmduh-483:
@moshe010 @adrianchiris

PR #70, opened 2020-04-26
Support GUID CNI runtime config

adrianchiris:
we should first:

1. Merge infiniband-guid as runtime config in multi-net-spec
2.  Update `k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1/types.go` with the new infiniband-guid runtime config. [PR 25](https://github.com/k8snetworkplumbingwg/network-attachment-definition-client/pull/25)
3. update PR go.mod to use a newer version of net-attach-def client from 2


adrianchiris:
All Prereq have been met.

This PR should ideally be split into 2:
- first 2 commits as general improvements
- last 2 commits for ib guid support

adrianchiris:
The change was validated and confirmed to be working.

README.md
```diff
@@ -87,6 +87,7 @@ metadata:
   namespace: kube-system
 data:
   DAEMON_SM_PLUGIN: "ufm" # Name of the subnet manager plugin
+  DAEMON_SM_PLUGIN_PATH: "/plugins" # Path to SM plugins folder
```
moshe010, line 90: where do you use it? I still see envDefault:"/plugins"

pkg/daemon/daemon.go
```diff
@@ -246,7 +248,9 @@ func (d *daemon) AddPeriodicUpdate() {
 					d.guidPodNetworkMap[allocatedGUID] = podNetworkID
 				}
 
-				if err = utils.SetPodNetworkGUID(network, allocatedGUID); err != nil {
+				// TODO(adrianc): should we just set GUID both in CNI ARGS and runtime config ?
```
moshe010, line 251: why it should be both?

pkg/daemon/daemon.go
```diff
@@ -263,11 +267,12 @@ func (d *daemon) AddPeriodicUpdate() {
 				pod.Annotations[v1.NetworkAttachmentAnnot] = string(netAnnotations)
 			}
 
-			// used GUID as net.HardwareAddress to use it in sm plugin which receive n[]et.HardwareAddress as parameter
+			// Create a GUID list as net.HardwareAddress to be used it in sm plugin
```
moshe010, line 270: I think you can remove the "it"

pkg/daemon/daemon.go
```diff
@@ -263,11 +267,12 @@ func (d *daemon) AddPeriodicUpdate() {
...
 			guidList = append(guidList, guidAddr.HardWareAddress())
 			passedPods = append(passedPods, pod)
 		}
 
+		// Get configured PKEY for network and Add the relevand POD GUIDs as members of the PKey via Subnet Manager
```
moshe010, line 275: s/Add/add
s/relevand/relevant

README.md
adrianchiris: This provides means to override the default plugins path.
its shown explicitly in the README and points to the plugins location inside the container image.
as seen [here](https://github.com/Mellanox/ib-kubernetes/blob/c82060e00b9b2e405ba1fc4a6886c59a5789372d/Dockerfile#L15).

This makes life easier when doing development, where i can just point this to my plugin build dir location.
(when not building image or deploying through k8s but invoking binary directly instead)

pkg/daemon/daemon.go
adrianchiris: will remove sure

README.md
moshe010: so why not move "/plugins" to constant as you used it in envDefault as well?

pkg/daemon/daemon.go
adrianchiris: i was thinking at the time of a cluster with different versions of ib-sriov cni where some might support infinibandGUID runtime config and some might not, however we should not hide configuration issues here so ill remove this TODO.

README.md
adrianchiris: This is a README , what do you mean by constant ?
in the code its listed once.

adrianchiris:
I will also open an issue to deprecate the GUID attribute in ib-sriov-cni so we use only one form

pkg/daemon/daemon.go
adrianchiris: Done

README.md
moshe010: I am talking about pkg/config/config.go  line 19

adrianchiris: correction to my above response, i dont think its possible to use variables as input to envDefault
as its a string annotation that `caarlos0/env/v6` use

https://github.com/caarlos0/env#supported-types-and-defaults

moshe010: you are right

PR #78, opened 2020-07-27
PeriodicAddUpdates and PeriodicDeleteUpdates code refactoring

DmytroLinkin:
Move code from Add and Delete periodic updates into dedicated
functions to remove code duplication, make code readable and get
an opportunity to cover functionality with unit tests in future.


adrianchiris:
this also refactors PeriodicDelete right ?

lets have two commits, one for update and one for delete

pkg/daemon/daemon.go
```diff
@@ -149,19 +157,128 @@ func (d *daemon) GetIbSriovNetwork(networkID string) (string, *utils.IbSriovCniS
 	return networkName, ibCniSpec, nil
 }
 
+// Get Pod's IB network specified by name. Update netMap with Pod's networks.
+func getIbSriovPodNetwork(netName string, pw *PodWrapper, netMap map[types.UID][]*v1.NetworkSelectionElement) error {
```
adrianchiris, line 161: in this method its not clear what is the input and what is the output.
also not too keen on the PodWrapper where it being used for both input and output

does it make sense just to update the pod network map here ? (and rename the method ofc)
main flow can just use : utils.GetPodNetwork()

pkg/daemon/daemon.go
```diff
@@ -149,19 +157,128 @@ func (d *daemon) GetIbSriovNetwork(networkID string) (string, *utils.IbSriovCniS
...
+		}
+
+		netMap[pw.pod.UID] = networks
+	}
+	network, err := utils.GetPodNetwork(networks, netName)
+	if err != nil {
+		return fmt.Errorf("failed to get pod networkName spec %s with error: %v", netName, err)
+	}
+
+	pw.ibNetwork = network
```
adrianchiris, line 178: what ensures that the retrieved network is indeed infiniband network ?

pkg/daemon/daemon.go
```diff
@@ -149,19 +157,128 @@ func (d *daemon) GetIbSriovNetwork(networkID string) (string, *utils.IbSriovCniS
 	return networkName, ibCniSpec, nil
 }
 
+// Get Pod's IB network specified by name. Update netMap with Pod's networks.
+func getIbSriovPodNetwork(netName string, pw *PodWrapper, netMap map[types.UID][]*v1.NetworkSelectionElement) error {
+	var err error
+	networks, ok := netMap[pw.pod.UID]
+	if !ok {
+		networks, err = netAttUtils.ParsePodNetworkAnnotation(pw.pod)
```
adrianchiris, line 165: I assume that first time a pod is processes we hit this code ?

pkg/daemon/daemon.go
```diff
@@ -149,19 +157,128 @@ func (d *daemon) GetIbSriovNetwork(networkID string) (string, *utils.IbSriovCniS
...
+		return fmt.Errorf("failed to get pod networkName spec %s with error: %v", netName, err)
+	}
+
+	pw.ibNetwork = network
+	pw.networks = networks
+	return nil
+}
+
+// Verify if GUID already exist for given network ID and allocates new one if not
+func (d *daemon) mapGUIDToPodNetwork(allocatedGUID, podNetworkID string, podUID types.UID) error {
```
adrianchiris, line 184: WDYT about renaming method to: `allocatePodNetworkGUID` ?

pkg/daemon/daemon.go
```diff
@@ -149,19 +157,128 @@ func (d *daemon) GetIbSriovNetwork(networkID string) (string, *utils.IbSriovCniS
...
+	if _, exist := d.guidPodNetworkMap[allocatedGUID]; exist {
```
adrianchiris, line 185: replate _ with podNetworkID to save additional access to map in the line below

or better yet just:
```
if allocatedPodNetworkID, _ := d.guidPodNetworkMap[allocatedGUID]; allocatedPodNetworkID != podNetworkID {
    return fmt.Errorf("failed to allocate......", allocatedGUID, allocatedPodNetworkID)
} 
```

pkg/daemon/daemon.go
```diff
@@ -119,19 +127,158 @@ func (d *daemon) Run() {
...
+	} else if err := d.guidPool.AllocateGUID(allocatedGUID); err != nil {
+		return fmt.Errorf("failed to allocate GUID for pod ID %s, wit error: %v", podUID, err)
+	} else {
+		d.guidPodNetworkMap[allocatedGUID] = podNetworkID
+	}
+
+	return nil
+}
+
+// Allocate GUID if network don't have it and update Pod's networks annotation
```
moshe010, line 199: s/don't/doens't

pkg/daemon/daemon.go
```diff
@@ -119,19 +127,158 @@ func (d *daemon) Run() {
...
+func (d *daemon) ProceedNetworkGUID(pw *PodWrapper, networkID string) error {
```
moshe010, line 200: you mean "ProceedNetworkGUID" ?

pkg/daemon/daemon.go
```diff
@@ -119,19 +127,158 @@ func (d *daemon) Run() {
...
+
+	netAnnotations, err := json.Marshal(pw.networks)
+	if err != nil {
+		return fmt.Errorf("failed to dump networks %+v of pod into json with error: %v", pw.networks, err)
+	}
+
+	pw.pod.Annotations[v1.NetworkAttachmentAnnot] = string(netAnnotations)
+
+	if err := d.kubeClient.SetAnnotationsOnPod(pw.pod, pw.pod.Annotations); err != nil {
+		if !strings.Contains(strings.ToLower(err.Error()), "not found") {
```
moshe010, line 257: can we use error code and not relay on string "not found" that may chang

pkg/daemon/daemon.go
```diff
@@ -144,128 +291,41 @@ func (d *daemon) AddPeriodicUpdate() {
...
-		}
-		log.Debug().Msgf("networkName attachment spec %+v", networkSpec)
-
-		ibCniSpec, err := utils.GetIbSriovCniFromNetwork(networkSpec)
-		if err != nil {
-			addMap.UnSafeRemove(networkID)
-			log.Warn().Msgf("failed to get InfiniBand SR-IOV CNI spec from network attachment %+v, with error %v",
-				networkSpec, err)
-			// skip failed network
+			if strings.Contains(err.Error(), "SR-IOV CNI spec") {
```
moshe010, line 296: we can move that to GetIbSriovNetwork
Also why this check is in if err != nil Line 295

pkg/daemon/daemon.go
```diff
@@ -144,128 +291,41 @@ func (d *daemon) AddPeriodicUpdate() {
...
+				pod: pod,
 			}
-			network, err := utils.GetPodNetwork(networks, networkName)
+			log.Debug().Msgf("pod namespace %s name %s", pod.Namespace, pod.Name)
+			err := getIbSriovPodNetwork(networkName, &pw, podNetworksMap)
 			if err != nil {
 				failedPods = append(failedPods, pod)
-				log.Error().Msgf("failed to get pod networkName spec %s with error: %v", networkName, err)
-				// skip failed pod
+				log.Err(err)
```
moshe010, line 315: better to check to Warning

pkg/daemon/daemon.go
adrianchiris: a comment would be mighty helpful for posterity then :)

pkg/daemon/daemon.go
DmytroLinkin: > IMHO we need to move all the logic of the cache network to different class
> NetworkCache class

What do You mean under "logic of the cache network"? If map update, then "all logic" is currently this map update only.

> get(netName,pw *PodWrapper)
> if netMap[pod id] exist return the NetworkSelectionElement
> else extract it from the PodWrapper.

Wrapper is temporary variable in loop, it doesn't hold networks from previous main loop iteration.

pkg/daemon/daemon.go
DmytroLinkin: > There is a check before the call for this method daemon.go to check if it contains ib-sriov-cni. I agree it not clear. Maybe we should move the check to this method

Currently main flow do next:
* for every network
* if network not IB - continue
* for every pod of that network
* get annotation of that network from pod

and doing check in this method looks weird, especially if network, which before loop we check is IB, cannot became non-IB.
Better then leave comment or rename function.

pkg/daemon/daemon.go
DmytroLinkin: Yes?

moshe010: Sorry I mean Do you mean "ProcessNetworkGUID"?  the method name is not clear to me

pkg/daemon/daemon.go
```diff
@@ -119,19 +127,158 @@ func (d *daemon) Run() {
...
+	allocatedGUID, err := utils.GetPodNetworkGUID(pw.ibNetwork)
+	podNetworkID := string(pw.pod.UID) + networkID
+	if err == nil {
+		// User allocated guid manually or Pod's network was rescheduled
+		err = d.mapGUIDToPodNetwork(allocatedGUID, podNetworkID, pw.pod.UID)
+		if err != nil {
+			return err
+		}
+
+		guidAddr, err = guid.ParseGUID(allocatedGUID)
```
adrianchiris, line 211: I have a feeling this should be called before L#206

pkg/daemon/daemon.go
DmytroLinkin: Yes, "Process" sounds better.

pkg/daemon/daemon.go
DmytroLinkin: function GetIbSriovNetwork used for Delete periodic update as well, but remove network from map for Add only

pkg/daemon/daemon.go
moshe010: so I want to warp netMap map[types.UID] the in a class/struct and move the logic of checking the map or getting i from PodWrapper in it so it won't be part of the daemon struct

pkg/daemon/daemon.go
moshe010: make sense.  so rename it. Also let the method return the network and the and assigen the pw.ibNetwork = getPodNetwork(...). It wired that the method is called getPod but is not returning network

pkg/daemon/daemon.go
adrianchiris: k8s error have a "one word" descriptor for an error which these methods compare agains

pkg/daemon/daemon.go
```diff
@@ -356,70 +419,26 @@ func (d *daemon) DeletePeriodicUpdate() {
...
-				continue
-			}
-
-			allocatedGUID, netErr := utils.GetPodNetworkGUID(network)
-			if netErr != nil {
-				failedPods = append(failedPods, pod)
-				log.Err(netErr)
+			guidAddr, err = getPodGUIDForNetwork(pod, networkName)
+			if err != nil {
+				if !strings.Contains(err.Error(), "is not Infiniband configured") {
```
adrianchiris, line 435: I think we should have an error construct with error code in addition to error message.
i see `strings.Contains()` appear in several places

this is something that did not exist prior to this refactor.

pkg/daemon/daemon.go
```diff
@@ -356,70 +419,26 @@ func (d *daemon) DeletePeriodicUpdate() {
...
-			continue
-		}
-		log.Debug().Msgf("networkName attachment spec %+v", networkSpec)
-
-		ibCniSpec, err := utils.GetIbSriovCniFromNetwork(networkSpec)
-		if err != nil {
-			log.Warn().Msgf("failed to get InfiniBand SR-IOV CNI spec from network attachment %+v, with error: %v",
-				networkSpec, err)
-			// skip failed networks
+			log.Err(err)
```
adrianchiris, line 424: to print the message: `log.Err(err).Msg("")`

pkg/daemon/daemon.go
```diff
@@ -356,70 +419,26 @@ func (d *daemon) DeletePeriodicUpdate() {
...
-			allocatedGUID, netErr := utils.GetPodNetworkGUID(network)
-			if netErr != nil {
-				failedPods = append(failedPods, pod)
-				log.Err(netErr)
+			guidAddr, err = getPodGUIDForNetwork(pod, networkName)
+			if err != nil {
+				if !strings.Contains(err.Error(), "is not Infiniband configured") {
+					failedPods = append(failedPods, pod)
+				}
+				log.Err(err)
```
adrianchiris, line 438: log.Err(err).Msg("") or log.Err(err).Send()

pkg/daemon/daemon.go
```diff
@@ -119,19 +127,158 @@ func (d *daemon) Run() {
...
+	return nil
+}
+
 func (d *daemon) AddPeriodicUpdate() {
 	log.Info().Msgf("running periodic add update")
 	addMap, _ := d.watcher.GetHandler().GetResults()
 	addMap.Lock()
 	defer addMap.Unlock()
+	// Contains ALL pods' networks
 	podNetworksMap := map[types.UID][]*v1.NetworkSelectionElement{}
```
adrianchiris, line 280: in the refactored code, i dont see where this is being used ?
can it be removed ? 

i mean, at the begining of the call its empty and is only populated by `getIbSriovPodNetwork`
then according to the code, its being called one per pod, so you never really read from it, am i worng ?

DmytroLinkin: Actually no. It is readed at L#163 and at first time, since there is no mapping it'll be written at L#171. But map persistent over all main flow, so when next networkID and same pod will be processed, L#163 will return mapped networks.

adrianchiris: i see , so first call for each pod will get ->**all**<- networks for the pod.

we should do one pass to get all needed information and only then start processing the pods IMO.

pkg/daemon/daemon.go
DmytroLinkin: Second case wouldn't work. Lets just stick to replate.

pkg/daemon/daemon.go
DmytroLinkin: Here's what I got for this:
```
type networksMap struct {
        theMap map[types.UID][]*v1.NetworkSelectionElement
}

// Return networks mapped to the pod. If mapping not exist it is created
func (n *networksMap) getPodNetworks(pod *kapi.Pod) ([]*v1.NetworkSelectionElement, error) {
        var err error
        networks, ok := n.theMap[pod.UID]
        if !ok {
                networks, err = netAttUtils.ParsePodNetworkAnnotation(pod)
                if err != nil {
                        return nil, fmt.Errorf("failed to read pod networkName annotations pod namespace %s name %s, with error: %v",
                                pod.Namespace, pod.Name, err)
                }

                n.theMap[pod.UID] = networks
        }
        return networks, nil
}
```
And `getIbSriovPodNetwork` function changed to `getPodNetworkInfo` which return `podInfo` (exWrapper), but without GUID as was initially.

pkg/daemon/daemon.go
DmytroLinkin: I don't see a real difference where to place it. It's funny, in call of `mapGUIDToPodNetwork` there is nested call to.. `ParseGUID(allocatedGUID), so in no error cases it's called twice.

pkg/daemon/daemon.go
DmytroLinkin: Like this?
```
type errorObj struct {
    code       int
    instance error
}
```

pkg/daemon/daemon.go
DmytroLinkin: Does it really significant to have separate loop to collect all pods' networks and then other loop to iterate over same networks? Assuming network map handler, I posted in upper comment, logic gonna look smthg like:
```
loop:
    getPodInfo
    processPodInfo
```

pkg/daemon/daemon.go
adrianchiris: the difference is where it would fail in case of invalid GUID.
either in `guid_pool:guidPool.AllocateGUID` or catch it early.

guid pool API kinda odd as when generating a GUID it returns a `GUID` type (uint64) but to allocate/release it you would need to convert it to a string.

adrianchiris: also the order here may hint that an allocation may pass but parsing the guid may fail, which is not the case

pkg/daemon/daemon.go
adrianchiris: OK lets see how it looks once PR is updated

pkg/daemon/daemon.go
adrianchiris: you should create a new package `errors`

```
package errors

import "fmt"

const (
    UNKNOWN_ERROR = iota
    // add error codes here
)

type CodedError struct {
    message string
    code int
}

func (e *CodedError)Error() string {
    return e.message
}

func Errorf(code uint, msg, args...) error {
    return CodedError{message: fmt.Sprintf(msg, args), code: code}
}

func CodeFromError(err error) int {
    switch e := err.(type) {
    case CodedError:
        return e.code
    default:
        return UNKNOWN_ERROR
    }
}
```

i dont guarantee the above compiles or works :) but it should give you a general idea

DmytroLinkin: Ok. I'll put it in separate commit.

pkg/errors/errors.go
```diff
@@ -0,0 +1,44 @@
...
+	// Keep last.
+	ErrorMax
+)
+
+type codedError struct {
+	code int
+	text string
+}
+
+func (e codedError) Error() string {
```
adrianchiris, line 24: This should be pointer receiver  i.e (e *codedError)

pkg/errors/errors.go
```diff
@@ -0,0 +1,44 @@
...
+	custom, ok := e.(*codedError)
+	if !ok {
+		return NotCodedErrorType
+	}
+	return custom.code
+}
+
+// Create new codedError.
+// If code !(NotCodedErrorType < code < ErrorMax) use UnknownError.
+func Errorf(code int, format string, a ...interface{}) error {
```
adrianchiris, line 39: a -> args ? or fmtArgs ?

pkg/errors/errors.go
```diff
@@ -0,0 +1,44 @@
...
+	code int
+	text string
+}
+
+func (e codedError) Error() string {
+	return e.text
+}
+
+// Return error code value or NotCodedErrorType if variable isn't of type codedError
+func ErrorCode(e error) int {
```
adrianchiris, line 29: as the package is already named `errors` you can just name it `Code(e error)` or `GetCode(e error)`

pkg/errors/errors.go
```diff
@@ -0,0 +1,44 @@
+package errors
+
+import "fmt"
+
+const (
+	// Value for destinguishing non-codedError type.
+	// Not used by codedError.
+	NotCodedErrorType = iota - 1
```
adrianchiris, line 8: if they all suffix with `Error` , this one should do the same.

adrianchiris: Also Error codes should be split from the coded error object, as we may reuse it with different error codes in other porjects.

pkg/errors/errors.go
```diff
@@ -0,0 +1,44 @@
+package errors
```
adrianchiris, line 1: package `codederror` ?

pkg/errors/errors.go
```diff
@@ -0,0 +1,44 @@
...
+	if !ok {
+		return NotCodedErrorType
+	}
+	return custom.code
+}
+
+// Create new codedError.
+// If code !(NotCodedErrorType < code < ErrorMax) use UnknownError.
+func Errorf(code int, format string, a ...interface{}) error {
+	if code <= NotCodedErrorType || code >= ErrorMax {
```
adrianchiris, line 40: i don't think this should be enforced.
ideally this coded error object can handle an arbitrary set of error codes so it may be possible to import to different projects.

pkg/daemon/daemon.go
```diff
@@ -119,153 +148,194 @@ func (d *daemon) Run() {
...
+				addMap.UnSafeRemove(networkID)
+			}
+			log.Error().Msgf("%v", err)
 			continue
 		}
-		log.Debug().Msgf("CNI spec %+v", ibCniSpec)
 
 		var guidList []net.HardwareAddr
-		var passedPods []*kapi.Pod
+		var passedPodsInfo []*podInfo
```
adrianchiris, line 324: IMO you can keep this as was before, i.e `passedPods` or `passedPodInfos`  where the `s` at the end helps the reader remember its a list.

however if you want to keep the change im fine as well.

pkg/daemon/daemon.go
```diff
@@ -333,18 +382,42 @@ func (d *daemon) AddPeriodicUpdate() {
...
+func getPodGUIDForNetwork(pod *kapi.Pod, networkName string) (net.HardwareAddr, error) {
+	networks, netErr := netAttUtils.ParsePodNetworkAnnotation(pod)
+	if netErr != nil {
+		return nil, fmt.Errorf("failed to read pod networkName annotations pod namespace %s name %s, with error: %v",
+			pod.Namespace, pod.Name, netErr)
+	}
+
+	network, netErr := utils.GetPodNetwork(networks, networkName)
+	if netErr != nil {
+		return nil, fmt.Errorf("failed to get pod networkName spec %s with error: %v", networkName, netErr)
```
adrianchiris, line 395: > networkName spec %s

network spec for network %s

pkg/daemon/daemon.go
```diff
@@ -333,18 +382,42 @@ func (d *daemon) AddPeriodicUpdate() {
...
+	}
+
+	allocatedGUID, netErr := utils.GetPodNetworkGUID(network)
+	if netErr != nil {
+		return nil, netErr
+	}
+
+	guidAddr, guidErr := net.ParseMAC(allocatedGUID)
+	if guidErr != nil {
+		return nil, fmt.Errorf("failed to parse allocated pod with error: %v", guidErr)
```
adrianchiris, line 409: > failed to parse allocated pod with error:

failed to parse allocated Pod GUID, error:

pkg/daemon/daemon.go
```diff
@@ -356,70 +429,26 @@ func (d *daemon) DeletePeriodicUpdate() {
...
-				continue
-			}
-
-			allocatedGUID, netErr := utils.GetPodNetworkGUID(network)
-			if netErr != nil {
-				failedPods = append(failedPods, pod)
-				log.Err(netErr)
+			guidAddr, err = getPodGUIDForNetwork(pod, networkName)
+			if err != nil {
+				if errors.ErrorCode(err) != errors.NotValidError {
```
adrianchiris, line 445: the error code name is lacking description IMO

pkg/daemon/daemon.go
```diff
@@ -333,18 +382,42 @@ func (d *daemon) AddPeriodicUpdate() {
...
+			pod.Namespace, pod.Name, netErr)
+	}
+
+	network, netErr := utils.GetPodNetwork(networks, networkName)
+	if netErr != nil {
+		return nil, fmt.Errorf("failed to get pod networkName spec %s with error: %v", networkName, netErr)
+	}
+
+	if !utils.IsPodNetworkConfiguredWithInfiniBand(network) {
+		return nil, errors.Errorf(errors.NotValidError, "network %+v is not InfiniBand configured", network)
```
adrianchiris, line 399: So this will now emit an error log instead of a warning as before.
i think we are OK with that

pkg/errors/errors.go
DmytroLinkin: Did it in same manner as other printf-like functions
`func Sprintf(format string, a ...interface{}) string {`

pkg/errors/errors.go
DmytroLinkin: > if they all suffix with `Error` , this one should do the same.

As I described in comment, this is not actually error. This is why it doesn't have suffix and, IMHO, with suffix it'll be confusing what this "Error" mean.

pkg/errors/errors.go
DmytroLinkin: Mb

pkg/errors/errors.go
DmytroLinkin: This protects from passed wrong code. As an alternative, code can be defined as new type, but in this way it should be exported. What do You think about this?
I don't see any problem with using it in other project and defining new codes.

pkg/daemon/daemon.go
```diff
@@ -42,6 +43,34 @@ type daemon struct {
...
+// Temporary struct used to proceed pods' networks
+type podInfo struct {
+	pod       *kapi.Pod
+	ibNetwork *v1.NetworkSelectionElement
+	networks  []*v1.NetworkSelectionElement
+	addr      net.HardwareAddr // GUID allocated for ibNetwork and saved as net.HardwareAddr
+}
+
+type networksMap struct {
+	theMap map[types.UID][]*v1.NetworkSelectionElement
```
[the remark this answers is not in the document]
DmytroLinkin: Definitely not the `netMap`, 'cause it becomes some sort of recursion: structure `networksMap` which hold `netMap`. And in code variable named `netMap`, so it's like `netMap.netMap`.
`innerMap`? Mb.
BTW, what's wrong with `theMap`?)

pkg/daemon/daemon.go
```diff
@@ -119,153 +148,194 @@ func (d *daemon) Run() {
...
+	}
+
+	return nil
+}
+
+// Allocate network GUID and update Pod's networks annotation. Add GUID to the podInfo instance
+func (d *daemon) ProcessNetworkGUID(pi *podInfo, networkID string) error {
+	var guidAddr guid.GUID
+	allocatedGUID, err := utils.GetPodNetworkGUID(pi.ibNetwork)
+	podNetworkID := string(pi.pod.UID) + networkID
```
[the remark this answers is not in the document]
DmytroLinkin: > also there should be a `_` separator for uniformity.

You mean `UID + "_" + networkID`

pkg/errors/errors.go
adrianchiris: how about: `NoCodedError`  ? this reads as there is no coded error but is still suffixed with `Error` ?

pkg/errors/errors.go
adrianchiris: Mega bits ? :)

pkg/errors/errors.go
adrianchiris: There are no wrong codes :) its just what the application decides to use.
perhaps with the exclusion of a special value (e.g `iota-1`) to mark a not coded error

if i use it in another project by:
```
import "github.com/Mellanox/ib-kubernetes/pkg/errors"

const (
    MyFirstError=iota
    MySecondError
    ...
    MyHunderdError
    MyLastError
)

func foo() error {
    return errors.Errof(MyHunderError, "Whoops something went wrong")
}
```

it will not work with current implemntation right ?

pkg/daemon/daemon.go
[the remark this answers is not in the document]
adrianchiris: it carries no meaning.
please show me examples in proper community maintained projects where they use `the` as a variable prefix and you got my blessing :)

pkg/daemon/daemon.go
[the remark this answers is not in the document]
adrianchiris: yes same as networkID is composed of `namespace_networkName`

pkg/daemon/daemon.go
```diff
@@ -119,153 +148,194 @@ func (d *daemon) Run() {
...
+			return err
+		}
+
+		if err = utils.SetPodNetworkGUID(pi.ibNetwork, allocatedGUID); err != nil {
+			return fmt.Errorf("failed to set pod network guid with error: %v ", err)
+		}
+
+		// Update Pod's network annotation here, so if network will be rescheduled we wouldn't allocate it again
+		netAnnotations, err := json.Marshal(pi.networks)
+		if err != nil {
```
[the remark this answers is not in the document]
DmytroLinkin: Not really. Allocating just parse guid string and set `p.guidPoolMap[guidAddr] = true`. If error occur, pod will be rescheduled and next time flow'll go into first branch of outer `if`.

pkg/daemon/daemon.go
```diff
@@ -119,153 +148,194 @@ func (d *daemon) Run() {
...
+		pi.pod.Annotations[v1.NetworkAttachmentAnnot] = string(netAnnotations)
+	}
+
+	// used GUID as net.HardwareAddress to use it in sm plugin which receive []net.HardwareAddress as parameter
+	pi.addr = guidAddr.HardWareAddress()
+	return nil
+}
+
+// Update and set Pod's network annotation
+func (d *daemon) UpdatePodNetworkAnnotation(pi *podInfo, removedList *[]net.HardwareAddr) error {
```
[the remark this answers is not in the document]
DmytroLinkin: Slice append happens in this function.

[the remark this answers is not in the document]
DmytroLinkin: Forgot to update function description.

pkg/errors/errors.go
DmytroLinkin: > Mega bits ? :)

Interesting association) "Maybe" I mean.

pkg/errors/errors.go
DmytroLinkin: Yes, it reads as there is an error "no coded")

pkg/errors/errors.go
DmytroLinkin: > There are no wrong codes

What stops me to create error like this?
`return errors.Errorf(461254, "Error text")`

> it will not work with current implemntation right ?

Almost. Need to modify enums little bit. Like this:
```
const (
    NotCodedErrorType = iota - 1
    MyFirstError
    MySecondError
    ...
    MyHunderdError
    MyLastError
    ErrorMax
)
```

pkg/daemon/daemon.go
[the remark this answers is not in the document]
adrianchiris: I still don't understand why removed list is a pointer to a slice and not the slice itself

pkg/errors/errors.go
adrianchiris: > Almost. Need to modify enums little bit. Like this

modify them where ?? the constants are not being overriden

adrianchiris: >What stops me to create error like this? ....
not the codedError struct

pkg/daemon/daemon.go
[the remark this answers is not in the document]
DmytroLinkin: > it carries no meaning.

Actually networksMap type has no meaning other than just to be a wrapper to the map.

> please show me examples in proper community maintained projects where they use the as a variable prefix and you got my blessing :)

Challenge accepted :D 
https://github.com/torvalds/linux/blob/master/tools/perf/tests/bp_account.c#L22
https://github.com/torvalds/linux/blob/master/tools/perf/tests/perf-hooks.c#L19
or better
https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/util/runtime/runtime_test.go#L341
this one seems to be a winner)
https://github.com/kubernetes/kubernetes/blob/master/vendor/sigs.k8s.io/kustomize/pkg/transformers/config/factorycrd.go#L188

pkg/errors/errors.go
DmytroLinkin: > modify them where ?? the constants are not being overriden
> the coded error implementation would still use the constants as defined in `...ib-kubernetes/pkg/errors`

Ah sorry, I thought old enums will be dropped, not extended.

> not the codedError struct

Didn't get what do You mean.

pkg/daemon/daemon.go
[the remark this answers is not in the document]
DmytroLinkin: Inside function underlying array will be recreated, while outside slice still would point to old array.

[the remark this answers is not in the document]
adrianchiris: right ! append would replace that.
which is in contrast of simply updating values of the existing array.

pkg/errors/errors.go
adrianchiris: its not `Errorf` responsibility to enforce that.

pkg/errors/errors.go
adrianchiris: if we split the error codes from the implementation and let the consumer of this package define the codes then you will be left with just one constant you need which is the special value `NotCodedErrorType`

What do you think ?

pkg/daemon/daemon.go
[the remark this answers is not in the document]
adrianchiris: Blessing it is then :)

pkg/utils/utils.go
```diff
@@ -186,3 +186,7 @@ func ParseNetworkID(networkID string) (string, string, error) {
 func GenerateNetworkID(network *v1.NetworkSelectionElement) string {
 	return fmt.Sprintf("%s_%s", network.Namespace, network.Name)
 }
+
+func GeneratePodNetworkID(pod *kapi.Pod, networkID string) string {
+	return string(pod.UID) + "_" + networkID
```
adrianchiris, line 191: nit: this could have been implemented the same as the above function for uniformity, but its fine :)
perhaps with a common internal function `func appendWithToken()` anyway this is not needed.

pkg/errcode/errcode.go
```diff
@@ -0,0 +1,31 @@
+package errcode
```
adrianchiris, line 1: can you add some short documentation comments above the package so a consumer of this package will know what it is and how to use it ?

example : https://github.com/golang/go/blob/master/src/context/context.go

pkg/errcode/errcode.go
```diff
@@ -0,0 +1,31 @@
...
+type errCode struct {
+	code int
+	text string
+}
+
+func (e *errCode) Error() string {
+	return e.text
+}
+
+// Return error code value or NotErrCodeType if variable isn't of type errCode
```
adrianchiris, line 20: Function/Method description comments should start with the method name, e.g
(golang documentation convention)

`// GetCode returns error code value .....`

lets adjust for this package

adrianchiris:
minor doc nits for errorcode and it LGTM (Looks Good To Me)

adrianchiris:
BTW @DmytroLinkin please make sure after all of the review iterations we did, we are left with something that actually works :)

pkg/errcode/errcode.go
```diff
@@ -26,12 +53,12 @@ func GetCode(e error) int {
...
-// Create new errCode
+// NewErr creates new errCode with provided text.
 func NewErr(code int, text string) error {
 	return &errCode{code: code, text: text}
 }
 
-// Create new errCode with formated text.
+// Errorf creates new errCode with formated text.
 func Errorf(code int, format string, a ...interface{}) error {
 	return &errCode{code: code, text: fmt.Sprintf(format, a...)}
```
adrianchiris, line 63: nit: This could have been implemented as : 
```
return NewErr(code, fmt.Sprintf(format, a...))
```

adrianchiris: or just remove NewErr altogether if it has no real use.
as its the same as calling `Errorf(code, text)`

DmytroLinkin:
Last commit message for errors package should be used instead of original

Issue #65, opened 2020-04-13
Some failed pods in add periodic update may become running in the process


Mmduh-483:
If multus didn't find the network CRD that a pod requested, then multus just ignore that CRD without failing. for the following pod:
```
...
annotations:
  k8s.v1.cni.cncf.io/networks: [{"name": "my-network"}]
...
```
if the network CRD `my-network` is not created, then multus will ignore it, so the pod will be running, while it is stuck in the ib-kubernetes add periodic update, even if the network created later, and the pod get configured, it will be late since the pod is running.

Mmduh-483:
@moshe010 @adrianchiris

DmytroLinkin:
Pod's network annotation:
```
...
  annotations:
    k8s.v1.cni.cncf.io/networks: some-network
...
```
After creating pod it hangs at `ContainerCreating` step.
```
$ kubectl describe pod failed-pod
...
  Warning  FailedCreatePodSandBox  0s (x4 over 6s)    kubelet, hostname (combined from similar events): 
Failed to create pod sandbox: rpc error: code = Unknown desc = failed to set up sandbox container 
"a1cf080aa5630453c5044803c97578bfb5ab2e9b6585a9c067ed39cddce16f45" network for pod "failed-pod": 
networkPlugin cni failed to set up pod "failed-pod_default" network: Multus: [default/failed-pod]: error loading 
k8s delegates k8s args: TryLoadPodDelegates: error in getting k8s network from pod: GetNetworkDelegates: 
failed getting the delegate: getKubernetesDelegate: cannot find get a network-attachment-definition
 (some-network) in namespace (default): network-attachment-definitions.k8s.cni.cncf.io "some-network" not found
```

adrianchiris:
I dont quite understand the issue @Mmduh-483,
The problem was with multus according to the original description, am i right ?

@DmytroLinkin please try the alternate form for setting the network annotation and see if it behaves the same.
i.e `k8s.v1.cni.cncf.io/networks: [{"name": "some-network"}]`

DmytroLinkin:
> @DmytroLinkin please try the alternate form for setting the network annotation and see if it behaves the same.
> i.e `k8s.v1.cni.cncf.io/networks: [{"name": "some-network"}]`

Creating pod with such annotation lead to
```
[...]

PR #82, opened 2020-08-05
Periodic update network

DmytroLinkin:
Some failed pods in add periodic update may become running in the process #65

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -19,7 +19,6 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 
 	"github.com/Mellanox/ib-kubernetes/pkg/config"
-	"github.com/Mellanox/ib-kubernetes/pkg/errcode"
 	"github.com/Mellanox/ib-kubernetes/pkg/guid"
 	k8sClient "github.com/Mellanox/ib-kubernetes/pkg/k8s-client"
 	"github.com/Mellanox/ib-kubernetes/pkg/sm"
@@ -181,8 +180,7 @@ func (d *daemon) getIbSriovNetwork(networkID string) (string, *utils.IbSriovCniS
 		}
 		return true, nil
 	}); err != nil {
-		return "", nil, errcode.Errorf(ErrGetNetAttachDef, "failed to get networkName attachment %s",
-			networkName)
+		return "", nil, fmt.Errorf("failed to get networkName attachment %s", networkName)
 	}
 	log.Debug().Msgf("networkName attachment %v", netAttInfo)
 
@@ -195,7 +193,7 @@ func (d *daemon) getIbSriovNetwork(networkID string) (string, *utils.IbSriovCniS
 
 	ibCniSpec, err := utils.GetIbSriovCniFromNetwork(networkSpec)
 	if err != nil {
-		return "", nil, errcode.Errorf(ErrNotIBSriovNetwork,
+		return "", nil, fmt.Errorf(
 			"failed to get InfiniBand SR-IOV CNI spec from network attachment %+v, with error %v",
 			networkSpec, err)
 	}
@@ -227,8 +225,8 @@ func getPodNetworkInfo(netName string, pod *kapi.Pod, netMap networksMap) (*podN
 func (d *daemon) allocatePodNetworkGUID(allocatedGUID, podNetworkID string, podUID types.UID) error {
 	if mappedID, exist := d.guidPodNetworkMap[allocatedGUID]; exist {
 		if podNetworkID != mappedID {
-			return errcode.Errorf(ErrGUIDAlreadyAllocated,
-				"failed to allocate requested guid %s, already allocated for %s", allocatedGUID, mappedID)
+			return fmt.Errorf("failed to allocate requested guid %s, already allocated for %s",
+				allocatedGUID, mappedID)
 		}
 	} else if err := d.guidPool.AllocateGUID(allocatedGUID); err != nil {
 		return fmt.Errorf("failed to allocate GUID for pod ID %s, wit error: %v", podUID, err)
@@ -353,33 +351,22 @@ func (d *daemon) AddPeriodicUpdate() {
 		log.Info().Msgf("processing network networkID %s", networkID)
 		networkName, ibCniSpec, err := d.getIbSriovNetwork(networkID)
 		if err != nil {
-			code := errcode.GetCode(err)
-			if code == ErrNotIBSriovNetwork || code == ErrGetNetAttachDef {
-				// Not an IB SR-IOV network or failed to get network attachment definition
-				addMap.UnSafeRemove(networkID)
-				log.Error().Msgf("droping network: %v", err)
-			} else {
-				log.Debug().Msgf("skipping network: %v", err)
-			}
+			addMap.UnSafeRemove(networkID)
+			log.Error().Msgf("droping network: %v", err)
 			continue
 		}
 
 		var guidList []net.HardwareAddr
 		var passedPods []*podNetworkInfo
-		var failedPods []*kapi.Pod
 		for _, pod := range pods {
 			log.Debug().Msgf("pod namespace %s name %s", pod.Namespace, pod.Name)
 			var pi *podNetworkInfo
 			pi, err = getPodNetworkInfo(networkName, pod, netMap)
 			if err != nil {
-				failedPods = append(failedPods, pod)
 				log.Error().Msgf("%v", err)
 				continue
 			}
-			if err := d.processNetworkGUID(networkName, ibCniSpec, pi); err != nil {
-				if errcode.GetCode(err) != ErrGUIDAlreadyAllocated {
-					failedPods = append(failedPods, pod)
-				}
+			if err = d.processNetworkGUID(networkName, ibCniSpec, pi); err != nil {
 				log.Error().Msgf("%v", err)
 				continue
 			}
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -93,6 +93,11 @@ data:
   GUID_POOL_RANGE_END: "02:FF:FF:FF:FF:FF:FF:FF" # The last guid in the pool
 ```
 
+### Multus configuration
+
+Note: If "kubeconfig" field is missed in 00-multus.conf then every pod creation will ignore network
```
adrianchiris, line 98: > Note: If "kubeconfig" field is missed in 00-multus.conf

Note: If `kubeconfig` field is missing in multus CNI configuration ...

pkg/daemon/daemon.go
```diff
@@ -55,6 +55,14 @@ type networksMap struct {
 	theMap map[types.UID][]*v1.NetworkSelectionElement
 }
 
+// Exponential backoff ~26 sec + 6 * <api call time>
+// NOTE: k8s client has built in exponential backoff, which ib-kubernetes don't use.
+// In case of increased backoff time might be KUBE_CLIENT_BACKOFF_BASE & KUBE_CLIENT_BACKOFF_DURATION
```
adrianchiris, line 60: Its not clear from the comment what the maximal backoff time would be.
is it the maximal value between the two ? or something else ? as we are performing exponential backoff of an exponential backoff method ?

pkg/daemon/daemon.go
```diff
@@ -107,8 +115,16 @@ func NewDaemon() (Daemon, error) {
 		return nil, err
 	}
 
-	if err := smClient.Validate(); err != nil {
-		return nil, err
+	// Try to validate if subnet manager is reachable in backoff loop
+	var validateErr error
```
adrianchiris, line 119: i think for the initial startup of the daemon, we don't need to do exponential backoff WDYT ?

pkg/daemon/daemon.go
```diff
@@ -155,9 +171,19 @@ func (d *daemon) getIbSriovNetwork(networkID string) (string, *utils.IbSriovCniS
...
+	if err = wait.ExponentialBackoff(backoffValues, func() (bool, error) {
+		netAttInfo, err = d.kubeClient.GetNetworkAttachmentDefinition(networkNamespace, networkName)
+		if err != nil {
+			log.Warn().Msgf("failed to get networkName attachment %s with error %v",
+				networkName, err)
+			return false, nil
+		}
+		return true, nil
+	}); err != nil {
+		return "", nil, errcode.Errorf(ErrGetNetAttachDef, "failed to get networkName attachment %s",
```
adrianchiris, line 185: should we also add the last error to the returned error for completeness ?

pkg/daemon/daemon.go
```diff
@@ -277,11 +303,20 @@ func (d *daemon) updatePodNetworkAnnotation(pi *podNetworkInfo, removedList *[]n
...
+			if kerrors.IsNotFound(err) {
+				return false, err
+			}
+			log.Warn().Msgf("failed to update pod annotations with err: %v", err)
+			return false, nil
 		}
 
+		return true, nil
+	}); err != nil {
+		log.Error().Msgf("failed to update pod annotations")
```
adrianchiris, line 319: should we also add the last error to the returned error for completeness ?

pkg/daemon/daemon.go
```diff
@@ -332,7 +370,8 @@ func (d *daemon) AddPeriodicUpdate() {
 		var failedPods []*kapi.Pod
 		for _, pod := range pods {
 			log.Debug().Msgf("pod namespace %s name %s", pod.Namespace, pod.Name)
-			pi, err := getPodNetworkInfo(networkName, pod, netMap)
+			var pi *podNetworkInfo
+			pi, err = getPodNetworkInfo(networkName, pod, netMap)
```
adrianchiris, line 372: why was this change needed ?

pkg/daemon/daemon.go
```diff
@@ -352,15 +391,23 @@ func (d *daemon) AddPeriodicUpdate() {
 
 		// Get configured PKEY for network and add the relevant POD GUIDs as members of the PKey via Subnet Manager
 		if ibCniSpec.PKey != "" && len(guidList) != 0 {
-			pKey, err := utils.ParsePKey(ibCniSpec.PKey)
+			var pKey int
+			pKey, err = utils.ParsePKey(ibCniSpec.PKey)
```
adrianchiris, line 389: same question as in L#374

pkg/daemon/daemon.go
```diff
@@ -502,10 +574,19 @@ func (d *daemon) DeletePeriodicUpdate() {
 //  initPool check the guids that are already allocated by the running pods
 func (d *daemon) initPool() error {
 	log.Info().Msg("Initializing GUID pool.")
-	pods, err := d.kubeClient.GetPods(kapi.NamespaceAll)
-	if err != nil {
-		err = fmt.Errorf("failed to get pods from kubernetes: %v", err)
-		log.Err(err)
+
+	// Try to get pod list from k8s client in backoff loop
```
adrianchiris, line 554: i think for the initial startup of the daemon, we don't need to do exponential backoff WDYT ?

pkg/daemon/daemon.go
```diff
@@ -352,46 +385,54 @@ func (d *daemon) AddPeriodicUpdate() {
...
 		// Update annotations for PODs that finished the previous steps successfully
 		var removedGUIDList []net.HardwareAddr
 		for _, pi := range passedPods {
-			err := d.updatePodNetworkAnnotation(pi, &removedGUIDList)
-			if err != nil {
-				failedPods = append(failedPods, pi.pod)
-				log.Error().Msgf("%v", err)
-				continue
-			}
+			d.updatePodNetworkAnnotation(pi, &removedGUIDList)
```
adrianchiris, line 412: can we keep the original updatePodNetworkAnnotation signature and return error, then log it here ? it seems clearer to me that way so the the main flow is "aware" of the failure

pkg/daemon/daemon.go
```diff
@@ -319,29 +353,28 @@ func (d *daemon) AddPeriodicUpdate() {
...
 		if err != nil {
-			if errcode.GetCode(err) == ErrNotIBSriovNetwork {
-				// Not an IB SR-IOV network
+			code := errcode.GetCode(err)
+			if code == ErrNotIBSriovNetwork || code == ErrGetNetAttachDef {
+				// Not an IB SR-IOV network or failed to get network attachment definition
 				addMap.UnSafeRemove(networkID)
+				log.Error().Msgf("droping network: %v", err)
+			} else {
+				log.Debug().Msgf("skipping network: %v", err)
```
adrianchiris, line 362: in this case we will retry this network in the next periodic update call right ?

pkg/daemon/daemon.go
DmytroLinkin: Not sure. What if daemon initialized before manager was configured/becomes reachable?

pkg/daemon/daemon.go
DmytroLinkin: We discussed it week(?) ago and decided that warning inside backoff loop is enough.

pkg/daemon/daemon.go
DmytroLinkin: same answer as previous

pkg/daemon/daemon.go
DmytroLinkin: Linter complained.

pkg/daemon/daemon.go
DmytroLinkin: same as previous

pkg/daemon/daemon.go
DmytroLinkin: Same reason as for subnet manager. Probably we can let ib-k8s gona fail here without backoff as on manager validation to notify user about configuration issues, but not sure how this will impact on UX.

pkg/daemon/daemon.go
DmytroLinkin: Ok, If U asking

pkg/daemon/daemon.go
DmytroLinkin: > in this case we will retry this network in the next periodic update call right ?

Right

> why would we want to do that ? in case of corrupt net-attach-def object ?

I assume for cases, when error wasn't persistent.

pkg/daemon/daemon.go
adrianchiris: I prefer we have it as i suggested

pkg/daemon/daemon.go
adrianchiris: OK lets keep it as is

pkg/daemon/daemon.go
adrianchiris: ok

pkg/daemon/daemon.go
adrianchiris: OK lets add warning to the log in case of failure when retrying like other cases

pkg/daemon/daemon.go
adrianchiris: i would suggest a reword:
 ```
In this case, there may be an increased execution time as <api call time> may take longer as retries are performed by k8s client.
k8s client backoff time is determined by KUBE_CLIENT_BACKOFF_BASE, KUBE_CLIENT_BACKOFF_DURATION environment variables [1]......
```

adrianchiris:
One concern about the delete not rescheduling is that it may leave the SM with stale entries if/when it comes back online.
In any case this type of cleaning should be handled in a separate routine.

Im LGTM on these changes.

PR #84, opened 2020-08-13
README: Add Multus configuration NOTE

DmytroLinkin:


README.md
```diff
@@ -93,6 +93,11 @@ data:
   GUID_POOL_RANGE_END: "02:FF:FF:FF:FF:FF:FF:FF" # The last guid in the pool
 ```
 
+### Multus configuration
```
adrianchiris, line 96: I dont think we need a new section, you could just add a note like below.
Also i propose a reword for the note, so its clearer to the _User_ the implication of a miss configuration in multus when `ib-kubernetes` is used.

>__Note__: For Infiniband workloads to work properly, multus CNI must be configured to work with kubernetes API by specifying the `kubeconfig` field in its configurations. If it is missing, then the Pod's infiniband network will not be properly set up.

adrianchiris: @DmytroLinkin  the formatting in my reply was intentional

PR #92, opened 2022-10-20
Avoid guid pool exhaustion from guid leaks by syncing it with SM

almaslennikov:
If the node running pods that use ib network is restarted, those pods'
 GUID are deleted from UFM but are persisted in ib-kubernetes.
They become unusable and might exhaust the GUID pool if it's configured to be small enough.
The solution is to sync guid pool with UFM because some GUIDs might have become free to use.

Signed-off-by: amaslennikov <amaslennikov@nvidia.com>

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -95,11 +95,6 @@ func NewDaemon() (Daemon, error) {
 		return nil, err
 	}
 
-	guidPool, err := guid.NewPool(&daemonConfig.GUIDPool)
-	if err != nil {
-		return nil, err
-	}
-
 	pluginLoader := sm.NewPluginLoader()
 	getSmClientFunc, err := pluginLoader.LoadPlugin(path.Join(
 		daemonConfig.PluginPath, daemonConfig.Plugin+".so"), sm.InitializePluginFunc)
@@ -124,6 +119,14 @@ func NewDaemon() (Daemon, error) {
 	}); err != nil {
 		return nil, validateErr
 	}
+
+	guidPool, err := guid.NewPool(&daemonConfig.GUIDPool)
+	if err != nil {
+		return nil, err
+	}
+
+	// Reset guid pool with already allocated guids to avoid collisions
+	err = syncGuidPool(smClient, guidPool)
 
 	podWatcher := watcher.NewWatcher(podEventHandler, client)
 	return &daemon{
@@ -256,7 +259,16 @@ func (d *daemon) processNetworkGUID(networkID string, spec *utils.IbSriovCniSpec
 	} else {
 		guidAddr, err = d.guidPool.GenerateGUID()
 		if err != nil {
-			return fmt.Errorf("failed to generate GUID for pod ID %s, with error: %v", pi.pod.UID, err)
+			switch err {
+			// If the guid pool is exhausted, need to sync with SM in case there are unsynced changes
+			case guid.GuidPoolExhaustedError:
+				err = syncGuidPool(d.smClient, d.guidPool)
+				if err != nil {
+					return err
+				}
+			default:
+				return fmt.Errorf("failed to generate GUID for pod ID %s, with error: %v", pi.pod.UID, err)
+			}
 		}
 
 		allocatedGUID = guidAddr.String()
@@ -281,6 +293,20 @@ func (d *daemon) processNetworkGUID(networkID string, spec *utils.IbSriovCniSpec
 
 	// used GUID as net.HardwareAddress to use it in sm plugin which receive []net.HardwareAddress as parameter
 	pi.addr = guidAddr.HardWareAddress()
+	return nil
+}
+
+func syncGuidPool(smClient plugins.SubnetManagerClient, guidPool guid.Pool) error {
+	usedGuids, err := smClient.ListGuidsInUse()
+	if err != nil {
+		return err
+	}
+
+	// Reset guid pool with already allocated guids to avoid collisions
+	err = guidPool.Reset(usedGuids)
+	if err != nil {
+		return err
+	}
 	return nil
 }
 
@@ -537,7 +563,7 @@ func (d *daemon) DeletePeriodicUpdate() {
 	log.Info().Msg("delete periodic update finished")
 }
 
-//  initPool check the guids that are already allocated by the running pods
+// initPool check the guids that are already allocated by the running pods
 func (d *daemon) initPool() error {
 	log.Info().Msg("Initializing GUID pool.")
 
--- a/pkg/guid/guid_pool.go
+++ b/pkg/guid/guid_pool.go
@@ -1,8 +1,8 @@
 package guid
 
 import (
+	"errors"
 	"fmt"
-
 	"github.com/rs/zerolog/log"
 
 	"github.com/Mellanox/ib-kubernetes/pkg/config"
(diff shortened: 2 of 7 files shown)


pkg/daemon/daemon.go
```diff
@@ -256,7 +267,22 @@ func (d *daemon) processNetworkGUID(networkID string, spec *utils.IbSriovCniSpec
 	} else {
 		guidAddr, err = d.guidPool.GenerateGUID()
 		if err != nil {
-			return fmt.Errorf("failed to generate GUID for pod ID %s, with error: %v", pi.pod.UID, err)
+			switch err {
+			// If the guid pool is exhausted, need to sync with SM in case there are unsynced changes
+			case guid.GuidPoolExhaustedError:
+				usedGuids, err := d.smClient.GetGuidsInUse()
```
e0ne, line 273: It's duplicated logic with lines #128-138. Let's move it into the separate function

pkg/guid/guid_pool.go
```diff
@@ -19,8 +20,14 @@ type Pool interface {
 	// ReleaseGUID release the reservation of the guid.
 	// It returns error if the guid is not in the range.
 	ReleaseGUID(string) error
+
+	// Reset clears the current pool and resets it with given values (may be empty)
+	Reset(guids []string) error
 }
 
+var GuidPoolExhaustedError = errors.New("GUID pool is exhausted")
```
e0ne, line 28: Should it be const?

e0ne (changes requested):

pkg/guid/guid_pool.go
almaslennikov: `errors.New` doesn't yield a const expression, so we can't do that :(

pkg/guid/guid_pool.go
```diff
@@ -50,6 +57,30 @@ func NewPool(conf *config.GUIDPoolConfig) (Pool, error) {
 	}, nil
 }
 
+func (p *guidPool) Reset(guids []string) error {
```
ykulazhenkov, line 60: Could you please add a docstring to the function?

pkg/guid/guid_pool.go
```diff
@@ -50,6 +57,30 @@ func NewPool(conf *config.GUIDPoolConfig) (Pool, error) {
...
+
+	p.guidPoolMap = map[GUID]bool{}
+	if guids == nil {
+		return nil
+	}
+
+	for _, guid := range guids {
+		err := p.AllocateGUID(guid)
+		if err != nil {
+			switch {
```
adrianchiris, line 71: alternative is to add a method: `guidInRange(guid) bool` and use it instead of swich case relying on error.

in fact most times you will have other guids configured in UFM.
my preference is to avoid relying on error of one method for a "good flow" of another.

that said, functionality wise it yields the same result so not a blocker imo

pkg/daemon/daemon.go
```diff
@@ -256,7 +259,16 @@ func (d *daemon) processNetworkGUID(networkID string, spec *utils.IbSriovCniSpec
 	} else {
 		guidAddr, err = d.guidPool.GenerateGUID()
 		if err != nil {
-			return fmt.Errorf("failed to generate GUID for pod ID %s, with error: %v", pi.pod.UID, err)
+			switch err {
+			// If the guid pool is exhausted, need to sync with SM in case there are unsynced changes
+			case guid.GuidPoolExhaustedError:
+				err = syncGuidPool(d.smClient, d.guidPool)
```
adrianchiris, line 265: is there a chance we end up in an in-consistent state here ?

im thinking in case DeletePeriodicUpdate is called at the same time.

almaslennikov: I'd say it doesn't really matter. `DeletePeriodicUpdate` doesn't introduce any new guids to the pool which could be deleted during a call to `Reset` and overwritten when a GUID for the next pod needs to be generated. Moreover, it makes an API call to UFM to release the GUID regardless of whether it is in the pool or not.
The model introduced with this fix offers eventual consistency for the GUID pool - so those changes would be reflect when the next sync occurs.
WDYT?

pkg/guid/guid_pool.go
almaslennikov: Thanks, changed it.

pkg/sm/plugins/ufm/ufm.go
```diff
@@ -133,6 +134,41 @@ func (u *ufmPlugin) RemoveGuidsFromPKey(pKey int, guids []net.HardwareAddr) erro
...
+}
+
+// GetGuidsInUse returns all guids currently in use by pKeys
+func (u *ufmPlugin) GetGuidsInUse() ([]string, error) {
+	response, err := u.client.Get(u.buildURL("/ufmRest/resources/pkeys/?guids_data=true"), http.StatusOK)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get the list of guids: %v", err)
+	}
+
+	var data map[string]interface{}
```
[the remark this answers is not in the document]
almaslennikov: I agree, it looks better - changed it.

pkg/sm/plugins/plugin.go
```diff
@@ -19,4 +19,7 @@ type SubnetManagerClient interface {
 	// RemoveGuidsFromPKey remove guids for given pkey.
 	// It return error if failed.
 	RemoveGuidsFromPKey(pkey int, guids []net.HardwareAddr) error
+
+	// GetGuidsInUse returns a list of all GUIDS associated with PKeys
+	GetGuidsInUse() ([]string, error)
```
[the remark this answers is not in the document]
almaslennikov: Renamed the method

pkg/sm/plugins/ufm/ufm.go
```diff
@@ -133,6 +134,41 @@ func (u *ufmPlugin) RemoveGuidsFromPKey(pKey int, guids []net.HardwareAddr) erro
 	return nil
 }
 
+// convertToMacAddr adds semicolons each 2 characters to convert to MAC format
+func convertToMacAddr(guid string) string {
```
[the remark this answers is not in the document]
almaslennikov: I'd say it's not needed - I would rely on UFM to do the necessary validation.
WDYT?

pkg/daemon/daemon.go
adrianchiris: ack so worst we will have some GUIDs in the pool that are not in use but will be marked as allocated.

pkg/daemon/daemon.go
```diff
@@ -125,6 +120,14 @@ func NewDaemon() (Daemon, error) {
 		return nil, validateErr
 	}
 
+	guidPool, err := guid.NewPool(&daemonConfig.GUIDPool)
+	if err != nil {
+		return nil, err
+	}
+
+	// Reset guid pool with already allocated guids to avoid collisions
```
adrianchiris, line 128: nit: Reset -> populate  since in this case the pool is empty.

pkg/sm/plugins/ufm/ufm.go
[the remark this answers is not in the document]
adrianchiris: ack, so perhaps add a comment on the expected string format returned from UFM WDYT ?

adrianchiris (approved):
minor nits otherwise LGTM

PR #102, opened 2024-09-05
Cleanup and maintenance

adrianchiris:


rollandf (approved):

pkg/drivers/http/client.go
```diff
@@ -35,11 +35,12 @@ func NewClient(isSecure bool, basicAuth *BasicAuth, cert string) (Client, error)
 	httpClient := &http.Client{Transport: http.DefaultTransport}
 	if isSecure {
 		if cert == "" {
-			/* #nosec */
+			//nolint:gosec
 			httpClient.Transport.(*http.Transport).TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
 		} else {
 			caCertPool := x509.NewCertPool()
 			caCertPool.AppendCertsFromPEM([]byte(cert))
+			////nolint:gosec
```
almaslennikov, line 43: Does the linter directive work with 4 slashes?

almaslennikov (approved):
Thank you for the effort! LGTM apart from a couple small comments

.golangci.yml
```diff
@@ -1,102 +1,146 @@
...
+  funlen:
+    lines: 120
+    statements: 55
+  goconst:
+    min-len: 2
+    min-occurrences: 2
+  gocyclo:
+    min-complexity: 30
+  goimports:
+    local-prefixes: github.com/Mellanox/maintenance-operator
```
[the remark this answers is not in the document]
adrianchiris: good catch !

pkg/drivers/http/client.go
adrianchiris: well it works :) but i will stick to 2.

PR #105, opened 2025-01-24
Setting GUID not randomly, but from a NetworkAttachmentDefinition "guid" field

punkerpunker:


punkerpunker:
Sorry, that's supposed to be a fork PR

PR #106, opened 2025-02-10
chore: add dependabot

rollandf:


rollandf:
Closing #104 already solves this

PR #167, opened 2025-09-04
Bugfixes

mladjan-gadzic:
- UFM Cleanup on Pod Lifecycle Events: Fixed issue where GUIDs weren't properly removed from UFM (Unified Fabric Manager) when pods completed (success/error) or were deleted.
- GUID Reallocation Conflicts: Added logic to remove existing GUID allocations from UFM before assigning a new partition (PKey) to prevent conflicts when the same GUID is reused.
- Pod State Handling: Improved pod lifecycle management by treating finished pods (succeeded/failed) the same as deleted pods for cleanup purposes.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -149,22 +149,18 @@ func NewDaemon() (Daemon, error) {
 		return nil, err
 	}
 
-	// Reset guid pool with already allocated guids to avoid collisions
-	err = syncGUIDPool(smClient, guidPool)
-	if err != nil {
-		return nil, err
-	}
-
 	podWatcher := watcher.NewWatcher(podEventHandler, client)
 	nadWatcher := watcher.NewWatcher(nadEventHandler, client)
+
+	// Return daemon fully formed
 	return &daemon{
 		config:            daemonConfig,
-		podWatcher:        podWatcher,
-		nadWatcher:        nadWatcher,
 		kubeClient:        client,
 		guidPool:          guidPool,
 		smClient:          smClient,
 		guidPodNetworkMap: make(map[string]string),
+		podWatcher:        podWatcher,
+		nadWatcher:        nadWatcher,
 		nadCache:          make(map[string]*v1.NetworkAttachmentDefinition),
 	}, nil
 }
@@ -278,8 +274,8 @@ func (d *daemon) becomeLeader() error {
 	log.Info().Msg("Becoming leader, initializing daemon logic")
 
 	// Initialize the GUID pool (rebuild state from existing pods)
-	if err := d.initPool(); err != nil {
-		log.Error().Msgf("initPool(): Leader could not init the guid pool: %v", err)
+	if err := d.initGUIDPool(); err != nil {
+		log.Error().Msgf("initGUIDPool(): Leader could not init the guid pool: %v", err)
 		return fmt.Errorf("failed to initialize GUID pool as leader: %v", err)
 	}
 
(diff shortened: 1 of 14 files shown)


almaslennikov (commented):
Hi @mladjan-gadzic, thank you for your contribution. Left some comments. Could you also cover the changed functionality with unit tests?

rollandf:
@mladjan-gadzic can you rebase and handle conflict?
Also maybe stash the commits to one?

pkg/daemon/daemon.go
```diff
@@ -546,8 +582,18 @@ func (d *daemon) DeletePeriodicUpdate() {
...
 			}
-
-			guidList = append(guidList, guidAddr)
+			podNetworkID := utils.GeneratePodNetworkID(pod, networkName)
+			if guidPodEntry, exist := d.guidPodNetworkMap[guidAddr.String()]; exist {
+				if podNetworkID == guidPodEntry {
+					log.Info().Msgf("matched guid %s to pod %s, removing", guidAddr, guidPodEntry)
+					guidList = append(guidList, guidAddr)
+				} else {
+					log.Warn().Msgf("guid %s is allocated to another pod %s not %s, not removing",
```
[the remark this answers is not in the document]
mladjan-gadzic: Added comment for this.

mladjan-gadzic:
@almaslennikov thanks for the review. i've pushed required changes. once you're okay with those, please resolve comments.

@rollandf thanks for the comment. i've squashed everything under one commit.

pkg/daemon/daemon.go
```diff
@@ -373,28 +385,84 @@ func (d *daemon) processNetworkGUID(
...
 	// Reset guid pool with already allocated guids to avoid collisions
-	err = guidPool.Reset(usedGuids)
+	err = d.guidPool.Reset(usedGuids)
 	if err != nil {
 		return err
 	}
+
+	// Remove stale GUIDs that are no longer in use by the subnet manager
+	// This handles cleanup of GUIDs from deleted/finished pods
+	for allocatedGUID, podNetworkID := range d.guidPodNetworkMap {
```
almaslennikov, line 440: At the time when syncGuidPool is called, the `d.guidPodNetworkMap` is empty because it's populated in the `d.initPool` method. Since we run both only once, let's combine them into a single function and properly handle the order of operations

almaslennikov (changes requested):

pkg/daemon/daemon.go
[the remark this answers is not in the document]
almaslennikov: Could you point me to the comment? I can't find it

mladjan-gadzic:
@almaslennikov, please, let me know if any additional changes are needed.

pkg/daemon/daemon.go
```diff
@@ -149,24 +149,30 @@ func NewDaemon() (Daemon, error) {
...
+		config:            daemonConfig,
+		watcher:           nil, // Will be set later
+		kubeClient:        client,
+		guidPool:          guidPool,
+		smClient:          smClient,
+		guidPodNetworkMap: make(map[string]string),
+	}
+
+	// Initialize guid pool with existing pods and sync with subnet manager
+	err = daemonInstance.initGUIDPool()
```
almaslennikov, line 163: I suggest to not run it here anymore since we only should start the actual logic if the instance becomes a leader. In the `becomeLeader` function we are already running `initGUIDPool`

almaslennikov: Let's refactor this function to return the daemon fully formed at the end

pkg/daemon/daemon.go
```diff
@@ -752,6 +805,7 @@ func (d *daemon) DeletePeriodicUpdate() {
 	log.Info().Msg("delete periodic update finished")
 }
 
+<<<<<<< HEAD
```
almaslennikov, line 808: There are merge artifacts left here

almaslennikov (changes requested):
@mladjan-gadzic could you test that the changes in this PR work? The current version doesn't build

pkg/daemon/daemon.go
mladjan-gadzic: done.

mladjan-gadzic:
> @mladjan-gadzic could you test that the changes in this PR work? The current version doesn't build

i've run make build/all/test/lint locally and it succeeds.

Issue #96, opened 2023-06-26
support HA  (leader election) in ib-kubernetes


adrianchiris:
k8s provides means to do that using leases. we should use it here as often done when implementing a k8s controller.

winsopc:
@adrianchiris  could you please review my changes to support this feature?
I tested in my dev cluster.

PR #168, opened 2025-09-04
Add High Availability support with Kubernetes leader election

winsopc:
Implement HA for ib-kubernetes using native Kubernetes leader election to prevent conflicts in multi-instance deployments.

- Add leader election with coordination.k8s.io/v1 Lease
- RBAC permissions for lease management
- Leader-only daemon operations (GUID/PKey management)
- Stable identity using Pod UID + node name
- Graceful failover and rolling update support

Closes: https://github.com/Mellanox/ib-kubernetes/issues/96


almaslennikov (approved):
Hi @winsopc the PR looks good after testing, LGTM!

PR #222, opened 2026-04-14
Fix GUID cleanup regression for multi-interface pods

winsopc:
Summary

  Fix inconsistent podNetworkID derivation introduced by #161 (Support multiple network interfaces per pod).

  Closes #207

  Problem

  processNetworkGUID() allocates and tracks GUID ownership using an interface-aware key (GeneratePodNetworkInterfaceID), but DeletePeriodicUpdate() matched using GeneratePodNetworkID and initGUIDPool() rebuilt entries using
  string(pod.UID) + network.Name.

  These identifiers are not equivalent for pods with multiple interfaces using the same NAD name, causing GUIDs to leak on delete and restart.

  Fix

  Introduce a shared GetPodNetworkInterfaceName helper that derives a stable per-interface name (uses InterfaceRequest when present, otherwise idx_N from occurrence index). All three paths now use this helper consistently:

  - add path (processNetworkGUID)
  - delete path (getAllPodGUIDInfosForNetwork)
  - init/restart path (initGUIDPool)

  Testing

  - Regression test confirms old code fails to match GUIDs for multi-interface pods
  - Init/restart consistency test verifies initGUIDPool + processNetworkGUID agree on IDs
  - GetPodNetworkInterfaceName unit tests for named and unnamed interface fallback

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -219,6 +219,34 @@ func GetAllPodNetworks(
 	return matchingNetworks, nil
 }
 
+// GetPodNetworkInterfaceName returns a stable per-interface name for a pod network.
+// When InterfaceRequest is empty, it falls back to the occurrence index among
+// networks with the same name so init, add, and delete all derive the same ID.
+func GetPodNetworkInterfaceName(
+	networks []*v1.NetworkSelectionElement, targetNetwork *v1.NetworkSelectionElement,
+) string {
+	if targetNetwork == nil {
+		return ""
+	}
+
+	if targetNetwork.InterfaceRequest != "" {
+		return targetNetwork.InterfaceRequest
+	}
+
+	sameNameIndex := 0
+	for _, network := range networks {
+		if network.Name != targetNetwork.Name {
+			continue
+		}
+		if network == targetNetwork {
+			return fmt.Sprintf("idx_%d", sameNameIndex)
+		}
+		sameNameIndex++
+	}
+
+	return "idx_0"
+}
+
 // ParsePKey returns parsed PKey from string
 func ParsePKey(pKey string) (int, error) {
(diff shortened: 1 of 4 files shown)


pkg/utils/utils.go
```diff
@@ -219,6 +219,34 @@ func GetAllPodNetworks(
...
+		return targetNetwork.InterfaceRequest
+	}
+
+	sameNameIndex := 0
+	for _, network := range networks {
+		if network.Name != targetNetwork.Name {
+			continue
+		}
+		if network == targetNetwork {
+			return fmt.Sprintf("idx_%d", sameNameIndex)
```
adrianchiris, line 242: will this work in case the same network is specified multiple times ?

e.g in the network annotations of the pod we got "ib-network-1" specified 2 times ?
i think in this case this function will be called twice but return the same interface name. (the first index where the network name matches)

winsopc: The function uses pointer identity (`network == targetNetwork`) not just name matching. Each call receives a different `*NetworkSelectionElement` pointer from the same parsed annotation slice, so the two `ib-network-1` entries produce `idx_0` and `idx_1` respectively. This is covered by the test at `utils_test.go:372`.

adrianchiris: ack. makes sense

rollandf:
@greptileai

pkg/utils/utils.go
```diff
@@ -219,6 +219,34 @@ func GetAllPodNetworks(
...
+		}
+		sameNameIndex++
+	}
+
```
rollandf, line 246: Maybe worth add a Warn log here

rollandf (approved):

almaslennikov (approved):

PR #221, opened 2026-04-13
feat: NICo plugin (NVIDIA NCX Infrastructure Controller) and supporting daemon/watcher changes

winsopc:
## Summary

Adds the **NICo plugin** — the in-cluster client for NVIDIA's NCX Infrastructure Controller — as a concrete `FabricClient` implementation, plus two supporting changes the NICo path needs:

- A small **daemon refactor** that snapshots periodic queues and extracts legacy plugin helpers so the partition-aware code path stays readable.
- A **watcher fallback** that drops back to `k8s.v1.cni.cncf.io/network-status` when `cni-args` is stripped from the pod, which happens on KubeVirt VMI delete because virt-handler removes the launcher pod's `cni-args` before the pod's final delete event reaches us.

NICo plugin uses the FabricClient hooks introduced in #228 to drive NAD-aware IB partition setup, partition readiness polling, and PF GUID retrieval through the public NICo REST SDK.

## Stack

This PR is **stacked on top of #228** (FabricClient interface). After #228 merges into upstream master and this branch is rebased, this PR will narrow to the 3 commits below.

PR #228, opened 2026-05-08
feat(daemon): add FabricClient support for IB partitions and PF GUIDs

winsopc:
## Summary

Adds an optional `FabricClient` interface that extends `SubnetManagerClient` with hooks for IB partition lifecycle (setup/teardown, readiness, NAD-driven cleanup) and node-level IB device discovery. No concrete implementation is included — plugins that don't implement the interface see no behavior change, since the daemon type-asserts at startup and falls back to the legacy GUID/PKey flow.

A reference consumer is the NICo plugin proposed in [#221](https://github.com/Mellanox/ib-kubernetes/pull/221). NICo (NVIDIA's NCX Infrastructure Controller) is a fabric manager whose responsibilities include IB partition lifecycle and exposing fabric inventory (partition state, node IB devices, factory-burned PF GUIDs, etc.) through a public REST API. The plugin uses the new hooks to surface those capabilities to the ib-kubernetes daemon.

## Scope

Targets the **fat-VM topology** used for GPU-IB AI workloads on KubeVirt — exactly one VMI per node per IB partition, where each VMI consumes that node's full set of IB PFs. Multi-pod-per-node-per-partition is an explicit non-goal here; lifting it requires a per-(networkID, nodeName) live-pod refcount and is tracked as a follow-up. The single-pod-per-(node, partition) assumption is documented inline in the relevant doc comments.

## Compatibility

- Plugins that do not implement `FabricClient` see no behavior change.
- The partition-aware code paths are unreachable without (a) a plugin that opts in via type assertion and (b) a NAD that opts in via `ibKubernetesEnabled`.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -18,24 +18,15 @@ package daemon
 
 import (
 	"context"
-	"encoding/json"
-	"errors"
 	"fmt"
-	"net"
 	"os"
 	"os/signal"
 	"path"
-	"sync"
 	"syscall"
 	"time"
 
-	v1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
-	netAttUtils "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/utils"
 	"github.com/rs/zerolog/log"
-	kapi "k8s.io/api/core/v1"
-	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/tools/leaderelection"
 	"k8s.io/client-go/tools/leaderelection/resourcelock"
@@ -45,7 +36,6 @@ import (
 	k8sClient "github.com/Mellanox/ib-kubernetes/pkg/k8s-client"
 	"github.com/Mellanox/ib-kubernetes/pkg/sm"
 	"github.com/Mellanox/ib-kubernetes/pkg/sm/plugins"
-	"github.com/Mellanox/ib-kubernetes/pkg/utils"
 	"github.com/Mellanox/ib-kubernetes/pkg/watcher"
 	resEvenHandler "github.com/Mellanox/ib-kubernetes/pkg/watcher/handler"
 )
@@ -55,36 +45,14 @@ type Daemon interface {
 	Run()
 }
 
+// daemon is the top-level wiring: config + plugin load + leader election +
+// the two controllers. All reconciliation lives in partitionController and
+// podController; daemon only orchestrates start/stop.
 type daemon struct {
-	config            config.DaemonConfig
-	podWatcher        watcher.Watcher
-	nadWatcher        watcher.Watcher // NAD watcher for network definition changes
-	kubeClient        k8sClient.Client
-	guidPool          guid.Pool
-	smClient          plugins.SubnetManagerClient
-	fabricClient      plugins.FabricClient // nil if plugin does not implement FabricClient
-	guidPodNetworkMap map[string]string    // allocated guid mapped to the pod and network
-
-	// NAD caches — sync.Map for concurrent access from AddPeriodicUpdate + ProcessNADChanges goroutines.
-	nadCache              sync.Map // network ID (string) -> *v1.NetworkAttachmentDefinition
-	partitionManagedCache sync.Map // network ID (string) -> bool
-}
-
-// Temporary struct used to proceed pods' networks
-type podNetworkInfo struct {
-	pod       *kapi.Pod
-	ibNetwork *v1.NetworkSelectionElement
-	networks  []*v1.NetworkSelectionElement
-	addr      net.HardwareAddr // GUID allocated for ibNetwork and saved as net.HardwareAddr
-}
-
-type podGUIDInfo struct {
-	addr         net.HardwareAddr
-	podNetworkID string
-}
-
-type networksMap struct {
-	theMap map[types.UID][]*v1.NetworkSelectionElement
+	config        config.DaemonConfig
+	kubeClient    k8sClient.Client
+	podCtrl       *podController
+	partitionCtrl *partitionController
 }
 
 // Exponential backoff ~26 sec + 6 * <api call time>
@@ -93,24 +61,10 @@ type networksMap struct {
 // NOTE: ufm client has default timeout on request operation for 30 seconds.
 var backoffValues = wait.Backoff{Duration: 1 * time.Second, Factor: 1.6, Jitter: 0.1, Steps: 6}
 
-// Return networks mapped to the pod. If mapping not exist it is created
-func (n *networksMap) getPodNetworks(pod *kapi.Pod) ([]*v1.NetworkSelectionElement, error) {
-	var err error
-	networks, ok := n.theMap[pod.UID]
-	if !ok {
-		networks, err = netAttUtils.ParsePodNetworkAnnotation(pod)
-		if err != nil {
-			return nil, fmt.Errorf("failed to read pod networkName annotations pod namespace %s name %s, with error: %v",
-				pod.Namespace, pod.Name, err)
-		}
-
-		n.theMap[pod.UID] = networks
-	}
-	return networks, nil
-}
-
-// NewDaemon initializes the need components including k8s client, subnet manager client plugins, and guid pool.
-// It returns error in case of failure.
+// NewDaemon initializes the daemon: config, k8s client, SM plugin, GUID pool,
+// pod + NAD watchers, and the two controllers. Construction is two-step so
+// the controllers can satisfy each other's dependencies before either
+// goroutine starts.
 func NewDaemon() (Daemon, error) {
 	daemonConfig := config.DaemonConfig{}
 	if err := daemonConfig.ReadConfig(); err != nil {
@@ -140,7 +94,8 @@ func NewDaemon() (Daemon, error) {
 		return nil, err
 	}
 
-	// Check if plugin implements FabricClient (optional interface for partition-aware plugins)
+	// Plugins that implement FabricClient drive partition-aware paths;
+	// others (UFM, noop) stay on the legacy GUID pool flow.
 	var fabricClient plugins.FabricClient
 	if pc, ok := smClient.(plugins.FabricClient); ok {
 		fabricClient = pc
@@ -168,16 +123,21 @@ func NewDaemon() (Daemon, error) {
 	podWatcher := watcher.NewWatcher(podEventHandler, client)
 	nadWatcher := watcher.NewWatcher(nadEventHandler, client)
 
-	// Return daemon fully formed
+	// Two-step wiring resolves the mutual dependency between controllers:
+	//   - partitionController needs podController.updatePodNetworkAnnotation
+	//     (passed by value as a method reference)
+	//   - podController needs partitionController as its PartitionReader
+	// Both controllers are constructed before any goroutine starts, so the
+	// post-construction pointer assignment is safe.
+	partitionCtrl := newPartitionController(fabricClient, smClient, client, nadWatcher)
+	podCtrl := newPodController(client, smClient, guidPool, podWatcher, partitionCtrl)
+	partitionCtrl.updatePodAnnotation = podCtrl.updatePodNetworkAnnotation
+
 	return &daemon{
-		config:            daemonConfig,
-		kubeClient:        client,
-		guidPool:          guidPool,
-		smClient:          smClient,
-		fabricClient:      fabricClient,
-		guidPodNetworkMap: make(map[string]string),
-		podWatcher:        podWatcher,
-		nadWatcher:        nadWatcher,
+		config:        daemonConfig,
+		kubeClient:    client,
+		podCtrl:       podCtrl,
+		partitionCtrl: partitionCtrl,
 	}, nil
 }
 
@@ -219,7 +179,6 @@ func (d *daemon) Run() {
 
 	log.Info().Msgf("Starting leader election in namespace: %s with identity: %s", namespace, identity)
 
-	// Create leader election configuration
 	lock := &resourcelock.LeaseLock{
 		LeaseMeta: metav1.ObjectMeta{
 			Name:      "ib-kubernetes-leader",
@@ -238,18 +197,17 @@ func (d *daemon) Run() {
 		RenewDeadline:   30 * time.Second, // Standard Kubernetes components deadline
 		RetryPeriod:     20 * time.Second, // Standard Kubernetes components retry
 		Callbacks: leaderelection.LeaderCallbacks{
-			OnStartedLeading: func(ctx context.Context) {
+			OnStartedLeading: func(_ context.Context) {
 				log.Info().Msgf("Started leading with identity: %s", identity)
 				if err := d.becomeLeader(); err != nil {
 					log.Error().Msgf("Failed to become leader: %v", err)
-					// Cancel context to gracefully release lease and exit
 					cancel()
 					return
 				}
 			},
 			OnStoppedLeading: func() {
 				log.Error().Msgf("Lost leadership unexpectedly, identity: %s", identity)
-				// Leadership lost unexpectedly - force immediate restart for clean state
+				// Force restart for clean state.
 				os.Exit(1)
 			},
 			OnNewLeader: func(leaderIdentity string) {
@@ -262,19 +220,16 @@ func (d *daemon) Run() {
 		},
 	}
 
-	// Start leader election in background
 	leaderElectionDone := make(chan struct{})
 	go func() {
 		defer close(leaderElectionDone)
 		leaderelection.RunOrDie(ctx, leaderElectionConfig)
 	}()
 
-	// Wait for termination signal or leader election completion
 	select {
 	case sig := <-sigChan:
 		log.Info().Msgf("Received signal %s. Terminating...", sig)
-		cancel() // This triggers ReleaseOnCancel
-		// Wait for graceful lease release
+		cancel() // releases lease via ReleaseOnCancel
 		select {
 		case <-leaderElectionDone:
 		case <-time.After(5 * time.Second):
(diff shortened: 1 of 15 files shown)


winsopc:
@rollandf PTAL.

rollandf:
@coderabbitai review

pkg/daemon/daemon.go
```diff
@@ -555,22 +634,23 @@ func (d *daemon) updatePodNetworkAnnotation(pi *podNetworkInfo, removedList *[]n
 
 		return true, nil
 	}); err != nil {
-		log.Error().Msgf("failed to update pod annotations")
+		log.Error().Err(err).Msg("failed to update pod annotations")
```
rollandf: Now, we can get here from `processPartitionPods`.
We should check if `pi.addr` is not nil before going to the `ReleaseGUID`

pkg/daemon/daemon.go
```diff
@@ -794,36 +915,350 @@ func (d *daemon) DeletePeriodicUpdate() {
...
+		return fmt.Errorf("failed to create IB partition: %v", err)
+	}
+
+	log.Info().Msgf("Created IB partition '%s' with PKey: %s for NAD %s/%s",
+		partitionName, partitionKey, nad.Namespace, nad.Name)
+
+	// Annotate NAD with PKey and add finalizer
+	if partitionKey != "" {
+		return d.updateNADPartitionKey(
+			fmt.Sprintf("%s_%s", nad.Namespace, nad.Name), partitionKey)
```
rollandf: we can reuse `partitionName` here, instead of rebuilding the same

pkg/daemon/daemon.go
```diff
@@ -346,6 +356,73 @@ func (d *daemon) getIbSriovNetwork(networkID string) (string, *utils.IbSriovCniS
...
+	}
+	enabled, ok := networkSpec[utils.IBKubernetesEnabled]
+	return ok && enabled == true
+}
+
+// deletePartitionPods removes a node's PF GUIDs from the partition for each deleted pod.
+// Assumes single-pod-per-node-per-partition (fat-VM); a per-(networkID,nodeName) refcount
+// over live pods would be needed before lifting that.
+func (d *daemon) deletePartitionPods(networkID string, pods []*kapi.Pod) {
+	partitionKey := d.getPartitionKeyFromNAD(networkID)
```
rollandf: It's only called when d.fabricClient != nil, but there's no in-function guard. A future refactor could accidentally call it directly. Consider adding an early-return check or a d.fabricClient assertion at the top.

pkg/daemon/daemon_test.go
```diff
@@ -716,4 +764,300 @@ var _ = Describe("Daemon", func() {
 			))
 		})
 	})
+
```
rollandf, line 837: Can we add tests also for:
processPartitionPods — especially the ErrPending retry path and the allProcessed tracking
deletePartitionPods — GUID extraction and RemoveGuidsFromPKey call
setupPartitionForNAD — annotation already present, non-ib-sriov skip, ibKubernetesEnabled gate
handleNADDeletion + removeNADFinalizer — retry-on-conflict logic

pkg/daemon/daemon.go
```diff
@@ -593,8 +673,41 @@ func (d *daemon) AddPeriodicUpdate() {
...
+			} else {
+				// Transient error — keep for next periodic run
+				log.Warn().Str("network_id", networkID).Err(err).Msg("NAD not ready, will retry")
+			}
+			continue
+		}
+
+		// Partition-managed NADs use hardware PF GUIDs from the FabricClient backend;
+		// non-partition NADs fall through to the legacy GUID/PKey flow below.
+		if d.fabricClient != nil && d.isPartitionManagedNAD(networkID) {
```
rollandf: maybe take this block out to a function? And remove the nolint funlen?

pkg/daemon/daemon.go
```diff
@@ -346,6 +356,73 @@ func (d *daemon) getIbSriovNetwork(networkID string) (string, *utils.IbSriovCniS
...
+// isPartitionManagedNAD checks if a NAD has ibKubernetesEnabled set to true,
+// meaning it should be managed via the partition-aware (FabricClient) path.
+func (d *daemon) isPartitionManagedNAD(networkID string) bool {
+	val, exists := d.nadCache.Load(networkID)
+	if !exists {
+		return false
+	}
+	nad := val.(*v1.NetworkAttachmentDefinition)
+	networkSpec := make(map[string]interface{})
+	if err := json.Unmarshal([]byte(nad.Spec.Config), &networkSpec); err != nil {
```
rollandf: Maybe worth to cache the value of `IBKubernetesEnabled` instead of unmarshall on every call?

winsopc:
@rollandf PTAL, I force-pushed the review update as `3bebc5b`.

Addressed the review comments:
- Added `pi.addr != nil` guard before `ReleaseGUID`.
- Reused `partitionName` instead of rebuilding it.
- Added in-function `fabricClient` guards.
- Extracted the partition add path into `processPartitionAddIfManaged` and removed the `funlen` nolint.
- Cached `ibKubernetesEnabled` via `partitionManagedCache`.
- Added tests for `processPartitionPods`, pending/allProcessed handling, `deletePartitionPods`, NAD setup gates, NAD deletion, and finalizer conflict retry.

rollandf (approved):
LGTM.

Let's get another review before merging

winsopc:
@almaslennikov PTAL when you have time.

pkg/watcher/handler/nad.go
```diff
@@ -74,15 +76,53 @@ func (n *NADEventHandler) OnAdd(obj interface{}, _ bool) {
...
+		n.deletedNADs.Set(networkID, newNAD)
+
+		log.Info().Msgf("Added terminating NAD to deletion queue: %s", networkID)
+	}
+}
+
+// OnDelete handles NAD deletion events (NAD fully removed from K8s)
+// Cleanup is handled in OnUpdate when DeletionTimestamp is set (Terminating state).
+// OnDelete fires AFTER finalizer is removed and NAD is fully gone.
+func (n *NADEventHandler) OnDelete(obj interface{}) {
```
almaslennikov, line 110: For a standalone deployment, probably need to add NAD to RBAC: https://github.com/Mellanox/ib-kubernetes/blob/master/deployment/ib-kubernetes.yaml#L7

pkg/daemon/daemon.go
```diff
@@ -60,10 +62,12 @@ type daemon struct {
...
 	guidPool          guid.Pool
 	smClient          plugins.SubnetManagerClient
-	guidPodNetworkMap map[string]string // allocated guid mapped to the pod and network
+	fabricClient      plugins.FabricClient // nil if plugin does not implement FabricClient
+	guidPodNetworkMap map[string]string    // allocated guid mapped to the pod and network
 
-	// NAD add-only cache
-	nadCache map[string]*v1.NetworkAttachmentDefinition // network ID -> NAD
+	// NAD caches — sync.Map for concurrent access from AddPeriodicUpdate + ProcessNADChanges goroutines.
+	nadCache              sync.Map // network ID (string) -> *v1.NetworkAttachmentDefinition
```
almaslennikov: We have two nadCaches now: one in the handler (written on OnAdd/OnDelete, exposed via GetNADFromCache) and one in the daemon (written via cacheNAD from the queue drain and from updateNAD* paths). Nothing reads the handler's cache. They have different eviction rules — handler evicts on OnDelete (after finalizer removal), daemon evicts in handleNADDeletion (before finalizer removal) — so during Terminating they disagree.

Can we pick one owner?

pkg/daemon/daemon.go
```diff
@@ -794,36 +957,354 @@ func (d *daemon) DeletePeriodicUpdate() {
...
+
+		d.cacheNAD(networkID, freshNAD)
+		return true, nil
+	})
+}
+
+// handleNADDeletion cleans up partition when NAD enters Terminating state.
+// The NAD still exists with its annotations — we clean up the backend partition,
+// then remove the finalizer so K8s can complete the deletion.
+func (d *daemon) handleNADDeletion(nad *v1.NetworkAttachmentDefinition) error {
```
almaslennikov: setupPartitionForNAD checks ib-sriov type and ibKubernetesEnabled before calling CreateIBPartition. The deletion side calls DeleteIBPartition for every ib-sriov NAD that enters Terminating, including those we never created a partition for. The interface says DeleteIBPartition must be idempotent, but pushing that contract onto plugin authors is unnecessary — we already know whether we own this NAD.

Either gate on d.isPartitionManagedNAD(networkID), or — cleaner — only act when our finalizer is present: if !hasNADFinalizer(nad, utils.PartitionNADFinalizer) { return nil }. The finalizer is our signal that we did the create.

pkg/daemon/daemon.go
```diff
@@ -794,36 +957,354 @@ func (d *daemon) DeletePeriodicUpdate() {
...
+			log.Info().Msgf("Successfully processed NAD delete event: %s", networkID)
+		}
+		deletedNADs.Unlock()
+	}
+
 	log.Debug().Msg("NAD changes processing completed")
 }
 
+// setupPartitionForNAD creates an IB partition for a NAD with ibKubernetesEnabled
+func (d *daemon) setupPartitionForNAD(nad *v1.NetworkAttachmentDefinition) error {
```
almaslennikov: Three different paths can call updateNADPartitionKey for the same NAD: setupPartitionForNAD (NAD add queue), processPartitionAddIfManaged (pod add path), and the backfill loop in ProcessNADChanges. They all run concurrently on different goroutines. Correctness is preserved by the conflict-retry inside the backoff, but you'll see noisy logs and wasted API calls.

almaslennikov (changes requested):
The feature shape is fine, but this PR adds ~540 lines to daemon.go and fuses a second reconciler into the daemon struct — pod/PKey and NAD/partition reconciliation now live in the same god-object, sharing two parallel caches with three code paths racing to write the PKey annotation.

The two loops are already decoupled by data flow: the NAD loop reads nothing from pod state, and the pod loop reads exactly three things from NAD state (cached NAD, ibKubernetesEnabled, PKey annotation). That's a clean producer/consumer split hiding inside one struct.

Before merging I'd like to see this extracted into two controllers:

PartitionController — owns the NAD informer, partition lifecycle, finalizer management, PKey backfill, NAD cache. Exposes a small read-only interface (LookupNAD, IsPartitionManaged, PartitionKey).
PodController — owns the pod informer, GUID pool, pod-to-PKey reconciliation. Takes the read-only NAD interface as a dependency.
daemon — shrinks to config, plugin load, leader election, signal handling, wiring. No reconciliation logic.
I'd also like each controller shaped so we can move it onto controller-runtime in a later PR without redesigning it. The pattern we're using (raw informers + SynchronizedMap work queues + wait.Until tickers + per-call backoff) is from the sample-controller era and it's the root cause of most of what makes this PR awkward — manual retry, periodic-tick latency, no per-key serialization, three goroutines racing on the same object. Not asking to switch frameworks now, but a few cheap choices now make the move trivial later:

Per controller, expose a Reconcile(ctx, key) (requeueAfter time.Duration, err error) method. Periodic ticks and the eventual workqueue both call the same entry point.
Move retry decisions into the return value (requeue / requeue-after / done), not into wait.ExponentialBackoff blocks inside the reconciler. The current ticker can interpret these the same way controller-runtime will.
Process one key at a time per controller (no parallel goroutines mutating the same NAD). This is what we'd get for free from a workqueue; doing it manually now eliminates the PKey-annotation race today.
Don't pass *daemon into reconcile helpers — pass the dependencies the controller actually needs. Makes them unit-testable without spinning up the whole daemon and means the eventual Reconciler struct already has its fields.
Side effects of the split itself: the two nadCache instances collapse into one (or we use the informer indexer that NewInformerWithOptions already returns), the PKey annotation gets a single writer, and create/delete gating becomes symmetric.

Two things to fix regardless of the refactor: the RBAC ClusterRole is missing update/patch on k8s.cni.cncf.io/* so every NAD update will be forbidden in a real cluster, and handleNADDeletion doesn't gate on isPartitionManagedNAD so it calls DeleteIBPartition for any ib-sriov NAD entering Terminating.

Happy to talk through the split.

pkg/watcher/handler/nad.go
winsopc: agree. I will fix it in next version

pkg/daemon/daemon.go
winsopc: Agreed.
 I’ll address this by making `partitionController` the single owner of NAD cache state.

The NAD handler should only queue add/delete events and should not keep a separate NAD cache or expose `GetNADFromCache`. Pod-side code will read NAD state through a narrow `PartitionReader.GetCachedNAD` interface owned by `partitionController`.

That should also make the Terminating behavior unambiguous: deletion and cache eviction will happen in one place during partition-controller NAD deletion processing, instead of having the handler and daemon evict at different points in the lifecycle.

I’ll update this and run the daemon/watcher tests before pushing the revision.

pkg/daemon/daemon.go
winsopc: I’ll use the finalizer as the ownership gate here.

`ibKubernetesEnabled` tells us the NAD is eligible for partition management, but the finalizer is the stronger signal that ib-kubernetes actually took ownership and attempted/finished the create path. So the delete path should skip backend cleanup unless `mellanox.infiniband.ib-kubernetes.io/partition` is present.

I’ll update the Terminating-NAD handling so it does:

if !hasNADFinalizer(nad, utils.PartitionNADFinalizer) {
    return nil
}

before calling `DeleteIBPartition`. That keeps `DeleteIBPartition` from needing to be idempotent for NADs this controller never owned.

I’ll also add/adjust tests for:
- ib-sriov NAD without our finalizer: no backend delete
- partition-managed NAD with our finalizer: backend delete + finalizer removal

winsopc:
@almaslennikov I add a new commit on top of first commit.  that could be easy to review.
PTAL when you have time.

rollandf:
@greptileai

pkg/daemon/partition_controller.go
```diff
@@ -0,0 +1,705 @@
...
+			}
+
+			if !hasPartitionFinalizer(entry.nad) {
+				p.deleteCachedNAD(entry.networkID)
+				deletedNADs.Remove(entry.networkID)
+				log.Debug().Str("network_id", entry.networkID).
+					Msg("NAD has no partition finalizer, nothing to clean up")
+				continue
+			}
+
```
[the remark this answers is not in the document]
winsopc: Fixed. The finalizer check now reads the NAD from nadCache (falling back to
the snapshot only on a cache miss) before both the hasPartitionFinalizer
gate and handleNADDeletion. The snapshot could predate a finalizer written
by a same-tick updateNADPartitionKey, since OnUpdate only reacts to the
DeletionTimestamp transition and never refreshes deletedNADs. Reading the
cache picks up the finalizer we just wrote, so the NAD is no longer stranded
in Terminating.

pkg/daemon/pod_controller.go
```diff
@@ -0,0 +1,760 @@
...
+				// attached to a partition we asked the backend to release.
+				// ErrPending also flows through here — keeps the entry until
+				// the backend confirms convergence to default partition.
+				log.Warn().Str("network_id", networkID).Err(err).
+					Msg("partition detach failed/pending, retrying on next tick")
+				continue
+			}
+			removeProcessedPodsFromQueue(deleteMap, networkID, pods)
+			continue
+		}
```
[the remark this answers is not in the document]
winsopc: Fixed. The IsPartitionManaged -> DeletePartitionPods branch now runs
before the getIbSriovNetwork lookup. Partition-managed detach reads the PKey
from the partition controller's cache and does not need the NAD spec, so a
NAD fetch failure (common during teardown, when the NAD is deleted alongside
its pods) no longer drops the delete-queue entry. The legacy path keeps its
original behavior.

pkg/daemon/partition_controller.go
```diff
@@ -0,0 +1,705 @@
...
+		// PF mode: only set mellanox.infiniband.app=configured (no pkey in cni-args).
+		// pi.addr is unset so the removedGUIDList sink stays empty.
+		var removedGUIDList []net.HardwareAddr
+		for _, pi := range podNetworkInfos {
+			if updateErr := p.updatePodAnnotation(pi, &removedGUIDList, ""); updateErr != nil {
+				log.Error().Err(updateErr).Msg("failed to update pod annotation")
+				allProcessed = false
+			}
+		}
```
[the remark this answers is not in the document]
winsopc: Fixed. Added a guard at the annotation call site in processPartitionPods:
if updatePodAnnotation is nil the pod is left unprocessed
(allProcessed=false, retried next tick) with an error log, instead of
panicking. Kept it at the call site rather than the top of the function so the
backend GUID attach still runs — the nil hook only blocks the annotation step.

pkg/daemon/pod_controller.go
```diff
@@ -0,0 +1,762 @@
...
+			deleteMap.Remove(networkID)
+			log.Warn().Msgf("droping network: %v", err)
+			continue
+		}
+
+		guidList := c.collectMatchedGUIDs(pods, networkName)
+		if err = c.removePKeyAndReleaseGUIDs(ibCniSpec, guidList); err != nil {
+			continue
+		}
+		removeProcessedPodsFromQueue(deleteMap, networkID, pods)
```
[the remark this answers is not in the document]
winsopc: Fixed. DeletePeriodicUpdate now warms the cache with GetCachedNAD before
the IsPartitionManaged check, so the first delete tick after start / leader
failover can't misroute a partition-managed detach to the legacy path. On a
fetch failure the check stays false and the legacy path retries, and a cache
hit still short-circuits the teardown case. Regression test added
(cold cache + API-only NAD → detach runs, entry dequeued).

pkg/daemon/partition_controller.go
```diff
@@ -0,0 +1,712 @@
...
+			if err := p.setupPartitionForNAD(entry.nad); err != nil {
+				log.Warn().Msgf("Failed to setup partition for NAD %s/%s: %v (will retry)",
+					entry.nad.Namespace, entry.nad.Name, err)
+				continue
+			}
+		}
+
+		log.Info().Msgf("Successfully processed NAD add event: %s", entry.networkID)
+		addedNADs.Remove(entry.networkID)
+	}
```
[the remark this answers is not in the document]
winsopc: Fixed. setupPartitionForNAD now re-fetches the NAD and returns early if it
is already Terminating, before CreateIBPartition — the add-queue entry is a
snapshot and can be stale. For the residual window (NAD enters Terminating
between that check and the finalizer write, which k8s then rejects), the
create + finalizer step is split so a failure routes to
rollbackPartitionIfTerminating: it re-checks the NAD and, if Terminating or
gone, deletes the just-created partition instead of orphaning it; otherwise it
returns the error so the add loop retries. Regression tests cover both the
skip-when-terminating and rollback paths.

pkg/daemon/partition_controller.go
```diff
@@ -0,0 +1,765 @@
...
+		partitionKey = pkey
+		if err := p.updateNADPartitionKey(networkID, partitionKey); err != nil {
+			log.Warn().Str("network_id", networkID).Err(err).
+				Msg("failed to persist PKey to NAD")
+		}
+	}
+
+	p.processPartitionPods(pods, networkName, partitionKey, netMap, addMap, networkID)
+	return true
+}
```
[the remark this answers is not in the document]
winsopc: Fixed. ProcessAddIfManaged now returns early (true, so the caller doesn't
fall through to the legacy path and the entry is retried next tick) when
updateNADPartitionKey fails, instead of proceeding to bind GUIDs. Binding
before the PKey annotation is persisted would leave the delete path unable to
derive the pkey (it reads the annotation), orphaning the GUIDs. This matches
the two sibling early-returns in the same block. Regression test added.

winsopc:
@rollandf @almaslennikov when you have time, please review my latest changes

pkg/daemon/partition_controller.go
```diff
@@ -0,0 +1,769 @@
...
+	}
+	if freshNAD.DeletionTimestamp != nil {
+		log.Info().Msgf("NAD %s/%s is terminating, skipping partition setup", nad.Namespace, nad.Name)
+		return nil
+	}
+
+	partitionName := networkIDForNAD(nad)
+	log.Info().Msgf("Creating IB partition for NAD %s/%s with name: %s", nad.Namespace, nad.Name, partitionName)
+
+	partitionKey, err := p.fabricClient.CreateIBPartition(partitionName)
```
almaslennikov, line 382: One more ordering concern: we create the backend partition before the NAD has our finalizer.

If the partition is created and then `updateNADPartitionKey` / `addNADFinalizer` fails for a transient API reason, the next tick will probably recover. But if the NAD is deleted before that recovery happens, the delete path currently treats the finalizer as the ownership signal and skips `DeleteIBPartition` when it is missing. Since `OnUpdate` also removes the NAD from the add queue once it enters Terminating, that partition can be left behind.

Could we add the finalizer first, then create the partition and annotate the PKey? `DeleteIBPartition` is documented as idempotent, so the delete path can safely handle the case where the finalizer exists but partition creation never completed.

pkg/daemon/partition_controller.go
```diff
@@ -0,0 +1,769 @@
...
+
+		if len(guids) == 0 {
+			log.Warn().Str("node", pod.Spec.NodeName).
+				Str("pod", pod.Name).Str("namespace", pod.Namespace).
+				Msg("no IB devices found on node")
+			allProcessed = false
+			continue
+		}
+
+		addErr := p.smClient.AddGuidsToPKey(pKey, guids)
```
almaslennikov, line 633: I think we still have a real leak here if the pod annotation write fails after `AddGuidsToPKey` succeeds.

In the legacy GUID path, `updatePodNetworkAnnotation` can release the allocated GUID on failure because `pi.addr` is set. In this PF path `pi.addr` is intentionally nil, so there is nothing for that helper to roll back. If `SetAnnotationsOnPod` fails because the pod was deleted, or the API write never lands, we keep the add queue entry but the node PF GUIDs are already attached to the tenant PKey. The delete path may never see a durable `configured` marker for that pod, so those GUIDs can stay in the partition.

Can we either roll back with `RemoveGuidsFromPKey(pKey, guids)` when the annotation update fails in this PF path, or change the ordering so the durable pod state is written before binding the PF GUIDs?

pkg/daemon/partition_controller.go
winsopc: Fixed — the ownership finalizer is now planted before CreateIBPartition, so a failed create or PKey write still cleans up through the idempotent delete path. Regression tests added.

pkg/daemon/partition_controller.go
winsopc: Fixed — on a pod-annotation failure after AddGuidsToPKey, the controller now rolls the GUIDs back with RemoveGuidsFromPKey and retries. Regression tests added.

almaslennikov (approved):
Thanks for addressing the comments, @winsopc! LGTM

winsopc:
Thanks @almaslennikov and @rollandf for the review and merge 🙏 Appreciate the time you both spent on this PR. I'll follow up with a Phase 2 PR building on this FabricClient/partition foundation — will open it as a draft for early feedback.

